PR #28, opened 2025-06-09
Feature/username password auth

StevenMPhillips:


serra:
@StevenMPhillips I was thinking about implementing sth similar, but IMO another approach would be better - curious about your thoughts.

From Dremio's perspective, I see two important downsides to this approach:
1/ Using usernames and passwords like this is not a good security practice in general and high quality software should not encourage bad practices.
2/ Creating personal access tokens for users is a differentiating feature of Dremio's enterprise edition.

To me, 1/ outweighs 2/ by a landslide, but I can imagine that in the current AI hype 2/ helps a bit in moving enterprise people who want leverage MCP to purchase an enterprise edition.

That being said, the community should be able to use this MCP server too.

My solution would be to keep the code to get a login token from the api outside this codebase. E.g. create a script that:
1/ retrieves a token from the api (valid for ~30hours iirc)
2/ starts the dremio service with this this token (command line parameter? docker mount? docker secret? env var?)
3/ restart before toke expiration (say daily)

It might be a bit hack-ish, but OK to me for community edition use.

And then the code here can stay secure and evolve with enterprise as well as community editions.

Thoughts?

PR #38, opened 2025-07-15
Change type annotations to match returned types, fixes #35

### LINKED ISSUE #35: Pydantic output validation errors for tools GetUsefulSystemTableNames and GetSchemaOfTable
**Context**

- Dremio OSS 24.3 running in Docker
- Supergateway http streamable proxy to stdio of Dremio MCP Server
- VS Code+CoPilot with Claude Sonnet 4.0 agent

**Affected Tools:**

- `GetUsefulSystemTableNames`
- `GetSchemaOfTable`

**Issue:** MCP validation error - server returning dict when list expected

**Error Message:**

```
1 validation error for invokeOutput
result
  Input should be a valid list [type=list_type, input_value={...}, input_type=dict]
```

**Root Cause:** The MCP server is returning JSON objects (`{}`) but the MCP protocol expects arrays (`[]`) for these tool responses.

**Impact:** Tools fail with validation errors despite containing correct data.

**Potential fixes:** Update server response formatting to wrap results in arrays or adjust MCP tool definitions to expect objects.

serra:
Fixes #35 Pydantic output validation errors for tools GetUsefulSystemTableNames and GetSchemaOfTable

### DIFF
--- a/tests/tools/test_output_validation.py
+++ b/tests/tools/test_output_validation.py
@@ -1,32 +1,43 @@
-"""
-Test script to demonstrate the actual MCP output validation bug.
-
-This script shows how FastMCP validates tool outputs by using the actual
-function type annotations, which is where the validation occurs in the MCP
-framework.
-"""
+#
+#  Copyright (C) 2017-2025 Dremio Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
 
 import pytest
-from typing import get_type_hints
-from pydantic import TypeAdapter
 from unittest.mock import patch
-
+from mcp.server.fastmcp.utilities.func_metadata import func_metadata
 from dremioai.tools.tools import GetUsefulSystemTableNames, GetSchemaOfTable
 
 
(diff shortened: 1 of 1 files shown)


serra:
I have some improvements coming for this today.

serra:
8f0fce1 is good to go.

pytest.ini
```diff
@@ -4,6 +4,7 @@ python_files = test_*.py
 python_classes = Test*
 python_functions = test_*
 asyncio_mode = strict
+asyncio_default_fixture_loop_scope = function
```
aniket-s-kulkarni, line 7: @serra - why do we need this for the test?

tests/tools/test_output_validation.py
```diff
@@ -0,0 +1,61 @@
+#
```
aniket-s-kulkarni, line 1: @serra I have a PR lined up for a larger change to the unit tests to validate the output types. Perhaps we can skip this test and I will follow up with that PR.

pytest.ini
serra: I set it beacause of this warning I got when running the test:

```
(dremioai) ➜  dremio-mcp git:(main) ✗ uv run pytest
/Users/marijn/dev/serraict/dremio-mcp/.venv/lib/python3.12/site-packages/pytest_asyncio/plugin.py:217: PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
=================================================================================== test session starts ====================================================================================
```

serra: But tbh I did not read into the details any further, so I'm open to any other suggestions.

serra: Reference: https://pytest-asyncio.readthedocs.io/en/stable/reference/configuration.html

PR #43, opened 2025-07-25
Recognize bearer auth tokens for streamable http

aniket-s-kulkarni:
- Adds basic framework for streamable http
- Adds framework to support bearer auth tokens, if specified, when in streamable http mode.


rdkworld:
Oh ok, Thanks for the clarification @aniket-s-kulkarni , will wait for complete functionality. I guess I will hold off on wrapper/proxy I was building over stdio server since that would be throw-away effort once streamable-http support is available

PR #68, opened 2025-09-25
Set URL and token for Dremio access

eickler:


eickler:
Sorry, that was an accident.

PR #99, opened 2026-04-09
DX-118683: Access tools from Dremio repository

AfonsoP1nto:
# Access tools from Dremio repository

## Summary
Enables the MCP server to proxy tool discovery and invocation to Dremio's Java-side tool registry via two new meta-tools, without duplicating tool definitions across repositories.

## Implementation
- New module \`ai_tools.py\` provides two async functions:
  - \`list_tools()\` — proxies \`GET /api/v4/ai/tools\`
  - \`invoke_tool(name, args)\` — proxies \`POST /api/v4/ai/tools/{name}:invoke\`, URL-encoding the tool name and wrapping args as \`{"args": ...}\`
- Two new MCP tools gated behind the \`enable_remote_tools\` setting:
  - \`DiscoverDynamicTools\` — calls \`list_tools()\` and returns the registry as JSON
  - \`CallDynamicTool\` — accepts a tool name and arguments (JSON string or dict) and proxies the invocation
- New \`requires_remote_tools\` class variable on \`Tools\` subclasses; \`is_tool_for()\` filters out tools carrying this flag when \`enable_remote_tools\` is false
- Configurable via \`enable_remote_tools\` setting under \`tools:\` (default: false)

## Features
- **Opt-in:** Both meta-tools are hidden unless \`enable_remote_tools: true\` is set in config
- **Soft-fail:** Errors from Dremio (unreachable, HTTP errors) are caught and returned as descriptive strings rather than crashing the tool invocation
- **Void tool handling:** Empty 200 responses (void tools) are explicitly distinguished from errors via \`InvokeToolResponse.is_empty\`
- **URL-safe tool names:** Tool names are percent-encoded before being interpolated into the invoke URL
- **Observability:** Both meta-tools are wrapped with \`@with_metrics\` for Prometheus tracking
- **Flexible arguments:** \`CallDynamicTool\` accepts \`tool_arguments\` as either a dict or a JSON string, with a clean error on invalid JSON

## Testing
- Unit tests for \`list_tools\` / \`invoke_tool\`: success, empty registry, HTTP error codes, URL encoding of special characters
- Pydantic model tests for \`AiTool\`, \`InvokeToolResponse\` (succeeded, failed, empty states)
- Integration tests for \`DiscoverDynamicTools\` and \`CallDynamicTool\`: registration gating, tool list proxying, error propagation, invalid JSON arguments
- Mock HTTP framework extended to support custom status codes and raise typed \`ClientResponseError\` on 4xx/5xx


src/dremioai/api/dremio/ai_tools.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import Dict, List, Optional, Any
+
+from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
+
+
+class AiTool(BaseModel):
+
+    name: str
+    description: Optional[str] = None
+    input_schema: Dict[str, Any] = Field(alias="inputSchema")
```
AfonsoP1nto: `_register_ai_tools()` has a fallback for missing `input_schema`, but `AiTool.input_schema` is currently required so `list_tools()` will raise if the API ever omits `inputSchema` (causing all remote tools to be skipped). Consider defaulting/making it optional to better handle older/partial API responses. (severity: medium, false-positive likelihood: 0.3)

src/dremioai/servers/mcp.py
```diff
@@ -258,9 +259,83 @@ async def _log_level_refresh_loop():
...
+            fn=handler,
+            name=name,
+            description=description,
+            parameters=input_schema,
+            fn_metadata=passthrough_metadata,
+            is_async=True,
+            context_kwarg=None,
+            annotations=ToolAnnotations(readOnlyHint=True),
+        )
+        mcp._tool_manager._tools[name] = tool_obj
```
AfonsoP1nto: Writing directly to `mcp._tool_manager._tools` relies on private FastMCP internals and may bypass any invariants/bookkeeping that the supported registration path performs. Is there a public API for registering a `Tool` to reduce upgrade/runtime risk? (severity: medium, false-positive likelihood: 0.4)

src/dremioai/servers/mcp.py
```diff
@@ -258,9 +259,83 @@ async def _log_level_refresh_loop():
...
+        registered += 1
+        _log.debug(f"Registered remote AI tool: {name}")
+
+    _log.info(f"Registered {registered} remote AI tool(s) from Dremio")
+
+
 @contextlib.asynccontextmanager
 async def _server_lifespan(app: FastMCP):
     """Lifespan context manager that runs background tasks alongside the server."""
+    await _register_ai_tools(app)
```
AfonsoP1nto: `await _register_ai_tools(app)` in the lifespan means server startup blocks on a network call to Dremio, which could noticeably delay or prevent startup if the endpoint is slow/unreachable. Consider ensuring startup still proceeds promptly even when tool discovery is degraded. (severity: medium, false-positive likelihood: 0.4)

AfonsoP1nto (commented):
Left a few high-signal notes around robustness/compatibility of remote AI tool registration.

src/dremioai/servers/mcp.py
```diff
@@ -294,9 +259,163 @@ async def _log_level_refresh_loop():
...
+                annotation=annotation,
+            )
+        )
+
+    # The actual implementation that proxies to invoke_tool
+    async def _handler_impl(**kwargs):
+        overrides = {}
+        try:
+            from mcp.server.auth.middleware.auth_context import get_access_token
+            from mcp.server.auth.provider import AccessToken
```
aniket-s-kulkarni: @AfonsoP1nto lets move all imports to top level.

src/dremioai/servers/mcp.py
```diff
@@ -294,9 +259,163 @@ async def _log_level_refresh_loop():
...
+    - Applies auth passthrough (PAT / project-id) and Prometheus metrics to
+      each proxied handler.
+
+    Note: tools are fetched **once** at server startup.  If the Java-side
+    registry changes (tools added / removed / signatures updated) the MCP
+    server must be restarted to pick up the changes.
+    """
+    _log = log.logger("remote_tools")
+    try:
+        remote_tools = await asyncio.wait_for(
```
aniket-s-kulkarni: This shouldn't be the case - MCP server may connect to old Dremio release or new or both at the same time as a part of the rollout. We cannot have this wait in here.

More importantly MCP clients look at tools once so as new releases add new MCP tools, those will never be refreshed by the clients.

We should reevaluate the approach 
1. have a tool - discover_dynamic_tools() that tells the LLM to call it to get information about more available tools from Dremio side
2. and then call_dynamic_tool() to actually do the call like above

aniket-s-kulkarni (changes requested):

src/dremioai/servers/mcp.py
```diff
@@ -267,6 +397,24 @@ def init(
             ),
         )
 
+    if settings.instance().tools and settings.instance().tools.enable_remote_tools:
+        mcp.add_tool(
+            make_logged_invoke(
+                "discover_dynamic_tools", _discover_dynamic_tools_handler
```
aniket-s-kulkarni: lets add this as a first class standard tool in tools.py .. no changes are required in mcp.py, they will auto picked up and registered.

aniket-s-kulkarni (changes requested):

src/dremioai/servers/mcp.py
```diff
@@ -13,8 +13,55 @@
...
+from mcp.types import ToolAnnotations
+from pydantic import AnyHttpUrl, field_serializer
+from pydantic.networks import AnyUrl
+
+from dremioai.metrics.registry import get_metrics_app
+from starlette.requests import Request
+from starlette.responses import Response
+
+
+class OAuthMetadataRFC8414(OAuthMetadata):
```
aniket-s-kulkarni: These changes feel like a rebase/merge snafu.

AfonsoP1nto: Yes, this was part of a merge conflict that was resolved

src/dremioai/tools/tools.py
```diff
@@ -537,6 +544,58 @@ async def invoke(self, query: str) -> Dict[str, Any]:
...
+        Call this tool to get a list of dynamically available tools with their
+        names, descriptions, and input schemas."""
+        try:
+            result = await ai_tools.list_tools()
+            return json.dumps(result)
+        except Exception as exc:
+            logger.warning(f"Failed to discover dynamic tools from Dremio: {exc}")
+            return (
+                f"Failed to discover dynamic tools from Dremio: {exc}. "
+                "The Dremio instance may be unreachable or may not support dynamic tools."
```
aniket-s-kulkarni: this statement should be removed. If dremio is unreachable then that is a separate ClientError that should be handled separately from catch all exception. In general the same MCP server should work with a coordinator that supports dynamic tools and another coordinator _of the same org and project_ that doesn't without revealing this. This situation will be temporary and eventually the old one will go away. A better way is to ask LLM to retry after some time

src/dremioai/tools/tools.py
```diff
@@ -537,6 +544,58 @@ async def invoke(self, query: str) -> Dict[str, Any]:
...
+    @with_metrics
+    async def invoke(self) -> str:
+        """Discover additional tools available from the Dremio server.
+        Call this tool to get a list of dynamically available tools with their
+        names, descriptions, and input schemas."""
+        try:
+            result = await ai_tools.list_tools()
+            return json.dumps(result)
+        except Exception as exc:
+            logger.warning(f"Failed to discover dynamic tools from Dremio: {exc}")
```
aniket-s-kulkarni: use `logger.exception()` and don't add `{exc}` to the string,

src/dremioai/tools/tools.py
```diff
@@ -537,6 +544,58 @@ async def invoke(self, query: str) -> Dict[str, Any]:
...
+            except json.JSONDecodeError as exc:
+                return f"Invalid JSON in tool_arguments: {exc}"
+
+        try:
+            result = await ai_tools.invoke_tool(tool_name, args)
+            if not result:
+                return json.dumps({"result": None})
+            return json.dumps(result)
+        except Exception as exc:
+            logger.warning(f"Failed to invoke dynamic tool '{tool_name}': {exc}")
```
aniket-s-kulkarni: Same as above - `logger.exception`.

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -0,0 +1,114 @@
...
+        raise AiToolError(
+            f"Failed to list AI tools: HTTP {e.status} {e.message}",
+            status=e.status,
+        ) from e
+    except Exception as e:
+        log.error(f"Failed to list AI tools: {e}")
+        raise AiToolError(f"Failed to list AI tools: {e}") from e
+
+
+async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> Dict[str, Any]:
```
aniket-s-kulkarni: change this to return -> `InvokeToolResponse`. Don't raise exceptions - catch them and turn them into an error message .. the .error attribute.

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -0,0 +1,114 @@
...
+    tools: List[AiTool]
+
+
+class InvokeToolResponse(BaseModel):
+
+    result: Optional[Any] = None
+    error: Optional[str] = None
+
+    @property
+    def succeeded(self) -> bool:
```
aniket-s-kulkarni: nit: instead of `def succeeded` use the more pythonic `def __nonzero__(self)` ... that will allow the callers to simply check this `if <response>`

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -0,0 +1,114 @@
...
+    name: str
+    description: Optional[str] = None
+    input_schema: Dict[str, Any] = Field(
+        default_factory=lambda: {"type": "object"}, alias="inputSchema"
+    )
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+
+class ListToolsResponse(BaseModel):
+    tools: List[AiTool]
```
aniket-s-kulkarni: consider adding an error field here as well and the __nonzero__ like below

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -0,0 +1,114 @@
...
+
+async def list_tools() -> List[Dict[str, Any]]:
+    try:
+        client = AsyncHttpClient()
+        response: ListToolsResponse = await client.get(
+            "/api/v4/ai/tools",
+            deser=ListToolsResponse,
+        )
+        return [t.model_dump(by_alias=False) for t in response.tools]
+    except ClientResponseError as e:
```
aniket-s-kulkarni, line 71: catch errors and return a ListToolsResponse with the appropriate error set in `.error`

src/dremioai/tools/tools.py
```diff
@@ -537,6 +544,58 @@ async def invoke(self, query: str) -> Dict[str, Any]:
         return {"results": res.to_dict(orient="records")}
 
 
+class DiscoverDynamicTools(Tools):
+    For: ClassVar[Annotated[ToolType, ToolType.FOR_SELF | ToolType.FOR_DATA_PATTERNS]]
+    requires_remote_tools: ClassVar[Annotated[bool, True]]
+
+    @secured
+    @with_metrics
+    async def invoke(self) -> str:
```
aniket-s-kulkarni, line 546: I suggest you redo it this way 

```python
# enabling dynamic tools per project is a better way to stage
# and test this functionality. Since listing of tools happens
# once at start of MCP server, it cannot be project scoped.
# So we have to check here

result = ListToolResult(
  result=None,
  error=None
         if settings.dremio.get("enable_remote_tools")
         else "Dynamic tools are disabled")

if result:
    try:
        result = await ai_tools.list_tools()
    except Exception as e:
        logger.exception(....)
        result.error = f"Failed to discover dynamic tools from Dremio: {exc}. "
 
return result.model_dump_json(...)
```

src/dremioai/tools/tools.py
```diff
@@ -537,6 +544,58 @@ async def invoke(self, query: str) -> Dict[str, Any]:
...
+            )
+
+
+class CallDynamicTool(Tools):
+    For: ClassVar[Annotated[ToolType, ToolType.FOR_SELF | ToolType.FOR_DATA_PATTERNS]]
+    requires_remote_tools: ClassVar[Annotated[bool, True]]
+
+    @secured
+    @with_metrics
+    async def invoke(self, tool_name: str, tool_arguments: Union[str, dict]) -> str:
```
aniket-s-kulkarni, line 561: The same as above

src/dremioai/config/settings.py
```diff
@@ -130,6 +130,10 @@ class Tools(BaseModel):
     server_mode: Annotated[
         Optional[Union[ToolType, int, str]], AfterValidator(_resolve_tools_settings)
     ] = Field(default=ToolType.FOR_SELF)
+    enable_remote_tools: Optional[bool] = Field(
```
aniket-s-kulkarni: Lets move this to `class Dremio` instead of here so it can be handled by Launch darkly.

aniket-s-kulkarni (commented):
@AfonsoP1nto this is looking pretty good. Just a few more comments and we can merge

src/dremioai/tools/tools.py
AfonsoP1nto: I've removed this. Since `list_tools()` now catches all exceptions internally and returns a `ListToolsResponse` with the `.error` field set, the tool itself is simplified to just `return result.model_dump_json()`. There is no error string to craft in the tool anymore.

src/dremioai/api/dremio/ai_tools.py
AfonsoP1nto: `invoke_tool()` now returns `InvokeToolResponse` in all cases. Both `ClientResponseError` and general exceptions are caught and turned into `InvokeToolResponse(error=...)`. I've also removed `AiToolError`.

src/dremioai/api/dremio/ai_tools.py
AfonsoP1nto: I've added `error: Optional[str] = None` to `ListToolsResponse` and a `__bool__` method (`return self.error is None`), matching the pattern on `InvokeToolResponse`.

src/dremioai/tools/tools.py
AfonsoP1nto: I've followed your suggestion. Since `list_tools()` now returns a `ListToolsResponse` (with `.error` set on failure, never raises), the invoke body is just:

```python
result = await ai_tools.list_tools()
return result.model_dump_json()
```

The `enable_remote_tools` guard is already enforced in `is_tool_for`, so these tools are never visible when the flag is off.

src/dremioai/tools/tools.py
```diff
@@ -225,6 +227,10 @@ def is_tool_for(
         if dremio is not None and dremio.project_id is None:
             return False
 
+    if get_requires_remote_tools(tool):
```
aniket-s-kulkarni: @AfonsoP1nto the changes look good, just one more - 

1. move `enable_remote_tools` to `settings.dremio` so we can use it via LD
2. move the check from here to the `invoke()` methods

aniket-s-kulkarni (approved):

PR #101, opened 2026-04-10
DX-118686: Address token expiry add Auth observability, tool error logging

aniket-s-kulkarni:
## Summary

- Add WARNING log in `RequireAuthWithWWWAuthenticateMiddleware.dispatch()` before returning 401 (logs request path + client host)
- Promote `"Token expired"` from DEBUG → WARNING in `JWKSVerifier.verify()` so auth failures are visible at production log level
- Apply 60-second expiry buffer in `DelegatingTokenVerifier.verify_token()` — tokens are rejected 60s before Auth0 expiry, giving clients a clean OAuth refresh window
- Add `_make_logged_invoke` decorator in `init()` to log tool invocation exceptions at WARNING with tool name + error detail before re-raising to MCP SDK
- Add INFO log in `verify_token()` when JWKS `verify()` returns `None` (graceful degradation path)

## Test Plan

- [ ] `TestTokenExpiryBuffer` (6 tests): buffer math for exp-59s, exp+61s, None exp, sentinel exp=0, no WARNING on happy path, INFO on JWKS degradation
- [ ] `TestDispatchWarning` (1 test): `dispatch()` emits WARNING with path + client on 401
- [ ] `TestMakeLoggedInvoke` (2 tests): wrapper logs WARNING on exception, no log on success
- [ ] Full suite: 315 passed, 0 failed

## JIRA

[DX-118686](https://dremio.atlassian.net/browse/DX-118686)

## Reviewer verdict

APPROVE WITH CHANGES — all reviewer feedback addressed (duplicate import fixed, missing tests for AC1 and AC2 added).

[DX-118686]: https://dremio.atlassian.net/browse/DX-118686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dremioai/servers/mcp.py
+++ b/src/dremioai/servers/mcp.py
@@ -171,8 +171,8 @@ class FastMCPServerWithAuthToken(FastMCP):
                         )
                         return None
                 else:
-                    self.logger.info(
-                        "JWKS verify() returned None — token expiry not enforced, forwarding to Dremio",
+                    self.logger.warning(
+                        "JWKS verify() returned None — rejecting token to force reauth",
                         project_id=ProjectIdMiddleware.get_project_id(),
                     )
                     return None
--- a/src/dremioai/servers/mcp.py
+++ b/src/dremioai/servers/mcp.py
@@ -13,57 +13,55 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-from mcp.server.auth.json_response import PydanticJSONResponse
-from mcp.server.fastmcp import FastMCP
-from mcp.server.fastmcp.prompts import Prompt
-from mcp.server.fastmcp.resources import FunctionResource
-from mcp.cli.claude import get_claude_config_path
-from mcp.types import ToolAnnotations
-from pydantic import AnyHttpUrl
-from pydantic.networks import AnyUrl
-
-from dremioai.metrics.registry import get_metrics_app
-from starlette.requests import Request
-from starlette.responses import Response
-
-from dremioai.api.oauth_metadata import OAuthMetadataRFC8414
-
-
-from dremioai.tools import tools
-import os
-from typing import List, Union, Annotated, Optional, Tuple, Dict, Any
-from functools import reduce, wraps
-from operator import ior
-from pathlib import Path
-from dremioai import log
-from typer import Typer, Option, Argument, BadParameter
-from rich import console, table, print as pp
-from click import Choice
-import logging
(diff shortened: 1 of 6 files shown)


ssaumitra (dismissed):
lgtm

src/dremioai/servers/mcp.py
```diff
@@ -88,6 +88,8 @@ async def dispatch(self, request: Request, call_next: RequestResponseEndpoint):
                 "Unauthorized request rejected",
```
ssaumitra, line 85: Can we also log on whether `request["user"]` is present or not? That way we will be able to know whether it was malformed request.

src/dremioai/servers/mcp.py
```diff
@@ -132,8 +144,15 @@ async def verify_token(self, token: str) -> AccessToken | None:
             is_verified = False
             if isinstance(self._jwks_verifier, JWKSVerifier):
                 if verified := await self._jwks_verifier.verify(token):
-                    expires_at, org_id = verified.exp, verified.aud
+                    expires_at = (verified.exp - _TOKEN_EXPIRY_BUFFER_SECONDS) if verified.exp is not None else None
+                    org_id = verified.aud
```
ssaumitra: After token expiry, can we send an error with hint response to LLM, asking it to renew the token pointing to the endpoint? Chatgpt says that it will help https://chatgpt.com/share/e/69d98409-21d4-800d-b248-4d08b4e94c69

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Good point, but actually this isn't the username at all. It's what we fill in `verify_token` as `client_id` parameter of the `AccessToken` return value - it's just "unused-client". We do not preserve the user name today. We shouldn't be logging user name or email in any case.

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Good point, but I am not sure it's the LLM that should be taking an action, it's the client/frontend that should. and HTTP/401 is pretty unambiguous response. The problem today is the error string contains "401" but the actual HTTP status code is `200`

ssaumitra:
The changes look good to me. I see that one test is failing though. Please let me know once that's fixed.

ssaumitra (approved):

PR #103, opened 2026-04-22
DX-118395: Make SearchTableAndViews resilient to broken catalog entries

ssaumitra:
A single unresolvable view in the search results (e.g. a view with a disconnected source) caused the catalog-by-path lookup to return HTTP 400, which propagated through asyncio.gather and failed the entire `SearchTableAndViews` tool call.

In query engine catalogs, it's a common behavior not to delete VDS when dependent PDS is deleted. However, when the schema of such VDS is queried, catalog throws an error. In this fix, we are choosing to be defensive when catalog does not return schema of certain dataset.

`get_schemas` now isolates per-path failures and returns the error message alongside each schema. `SearchTableAndViews` surfaces the affected paths in a `skipped` field plus a human-readable `skipped_note`, so the user sees the healthy results and a clear explanation for the ones that were dropped.

Also adds `CLAUDE.md` with project context and development guidelines to help Claude Code sessions work more effectively in this repo.

### DIFF
--- a/src/dremioai/api/dremio/catalog.py
+++ b/src/dremioai/api/dremio/catalog.py
@@ -18,18 +18,14 @@ from pydantic import BaseModel, Field, AfterValidator, ValidationError
 from typing import Annotated, List, Set, Tuple, Dict, AnyStr, Any, Union, Optional
 from datetime import datetime
 from enum import StrEnum, auto
-from functools import partial, reduce
-from csv import reader, excel
-from io import StringIO
-
-from aiohttp import ClientResponseError
+from functools import partial
 
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
 from dremioai.api.util import UStrEnum, run_in_parallel
 from dremioai.config import settings
-from dremioai import log
-
-logger = log.logger(__name__)
+from csv import reader, excel
+from io import StringIO
+from functools import reduce
 
 
 class CatalogItemType(UStrEnum):
@@ -190,34 +186,15 @@ async def get_schema(
     return schema
 
 
-class SchemaResult(BaseModel):
-    """Per-path result of `get_schemas`. `data` is populated on success; `error`
-    carries a human-readable failure message when the fetch raised."""
-
-    data: Optional[Dict[str, Any]] = None
-    error: Optional[str] = None
-
-
 async def get_schemas(
     dataset_path_or_ids: List[Union[List[str], str]],
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-    throw_exception: Optional[bool] = False,
-) -> SchemaResult:
-    async def _safe_get_schema(p) -> SchemaResult:
-        try:
-            data = await get_schema(p, by_id, include_tags, flatten)
-            return SchemaResult(data=data)
-        except Exception as e:
-            if throw_exception:
-                raise e
-            return SchemaResult(
-                error=f"{type(e).__name__}: {e}"
-            )
+) -> List[Dict[str, Any]]:
 
     return await run_in_parallel(
-        [_safe_get_schema(p) for p in dataset_path_or_ids]
+        [get_schema(p, by_id, include_tags, flatten) for p in dataset_path_or_ids]
     )
 
 
@@ -242,8 +219,7 @@ async def get_descriptions(
     components = set()
     result = {}
     while True:
-        results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True, throw_exception=True)
-        schemas = [r.data for r in results if r.data]
+        schemas = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
         rest = set()
         for s in schemas:
             if d := extract_description(s):
--- a/src/dremioai/api/dremio/catalog.py
+++ b/src/dremioai/api/dremio/catalog.py
@@ -20,6 +20,8 @@ from datetime import datetime
 from enum import StrEnum, auto
 from functools import partial
 
+from aiohttp import ClientResponseError
+
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
 from dremioai.api.util import UStrEnum, run_in_parallel
 from dremioai.config import settings
@@ -191,10 +193,17 @@ async def get_schemas(
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-) -> List[Dict[str, Any]]:
+) -> List[Tuple[Dict[str, Any], Optional[str]]]:
+    async def _safe_get_schema(p):
+        try:
+            return await get_schema(p, by_id, include_tags, flatten), None
+        except ClientResponseError as e:
+            return {}, f"HTTP {e.status}: {e.message}"
+        except Exception as e:
+            return {}, f"{type(e).__name__}: {e}"
 
     return await run_in_parallel(
-        [get_schema(p, by_id, include_tags, flatten) for p in dataset_path_or_ids]
+        [_safe_get_schema(p) for p in dataset_path_or_ids]
     )
 
 
@@ -219,7 +228,9 @@ async def get_descriptions(
     components = set()
     result = {}
     while True:
-        schemas = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
+        schemas = [
+            s for s, _ in await get_schemas(dataset_path_or_ids, by_id, include_tags=True) if s
+        ]
         rest = set()
         for s in schemas:
             if d := extract_description(s):
--- a/src/dremioai/api/dremio/catalog.py
+++ b/src/dremioai/api/dremio/catalog.py
@@ -204,7 +204,7 @@ async def get_schemas(
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
     throw_exception: Optional[bool] = False,
-) -> List[SchemaResult]:
+) -> SchemaResult:
     async def _safe_get_schema(p) -> SchemaResult:
         try:
             data = await get_schema(p, by_id, include_tags, flatten)
--- a/src/dremioai/api/dremio/catalog.py
+++ b/src/dremioai/api/dremio/catalog.py
@@ -18,16 +18,18 @@ from pydantic import BaseModel, Field, AfterValidator, ValidationError
 from typing import Annotated, List, Set, Tuple, Dict, AnyStr, Any, Union, Optional
 from datetime import datetime
 from enum import StrEnum, auto
-from functools import partial
+from functools import partial, reduce
+from csv import reader, excel
+from io import StringIO
 
 from aiohttp import ClientResponseError
 
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
 from dremioai.api.util import UStrEnum, run_in_parallel
 from dremioai.config import settings
-from csv import reader, excel
-from io import StringIO
-from functools import reduce
+from dremioai import log
+
+logger = log.logger(__name__)
 
 
 class CatalogItemType(UStrEnum):
@@ -188,19 +190,32 @@ async def get_schema(
     return schema
 
 
+class SchemaResult(BaseModel):
+    """Per-path result of `get_schemas`. `data` is populated on success; `error`
+    carries a human-readable failure message when the fetch raised."""
+
+    data: Optional[Dict[str, Any]] = None
+    error: Optional[str] = None
+
+
 async def get_schemas(
     dataset_path_or_ids: List[Union[List[str], str]],
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-) -> List[Tuple[Dict[str, Any], Optional[str]]]:
-    async def _safe_get_schema(p):
+) -> List[SchemaResult]:
+    async def _safe_get_schema(p) -> SchemaResult:
         try:
-            return await get_schema(p, by_id, include_tags, flatten), None
+            data = await get_schema(p, by_id, include_tags, flatten)
+            return SchemaResult(data=data)
         except ClientResponseError as e:
-            return {}, f"HTTP {e.status}: {e.message}"
+            return SchemaResult(
+                error=f"HTTP {e.status}: {e.message}"
+            )
         except Exception as e:
-            return {}, f"{type(e).__name__}: {e}"
+            return SchemaResult(
+                error=f"{type(e).__name__}: {e}"
+            )
 
     return await run_in_parallel(
         [_safe_get_schema(p) for p in dataset_path_or_ids]
@@ -228,9 +243,14 @@ async def get_descriptions(
     components = set()
     result = {}
     while True:
-        schemas = [
-            s for s, _ in await get_schemas(dataset_path_or_ids, by_id, include_tags=True) if s
-        ]
+        results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
+        for r in results:
+            if r.error is not None:
+                logger.info(
+                    "get_descriptions schema fetch failed",
+                    reason=r.error,
+                )
+        schemas = [r.data for r in results if r.data]
         rest = set()
         for s in schemas:
             if d := extract_description(s):
--- a/src/dremioai/api/dremio/catalog.py
+++ b/src/dremioai/api/dremio/catalog.py
@@ -203,16 +203,15 @@ async def get_schemas(
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
+    throw_exception: Optional[bool] = False,
 ) -> List[SchemaResult]:
     async def _safe_get_schema(p) -> SchemaResult:
         try:
             data = await get_schema(p, by_id, include_tags, flatten)
             return SchemaResult(data=data)
-        except ClientResponseError as e:
-            return SchemaResult(
-                error=f"HTTP {e.status}: {e.message}"
-            )
         except Exception as e:
+            if throw_exception:
+                raise e
             return SchemaResult(
                 error=f"{type(e).__name__}: {e}"
             )
@@ -243,13 +242,7 @@ async def get_descriptions(
     components = set()
     result = {}
     while True:
-        results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
-        for r in results:
-            if r.error is not None:
-                logger.info(
-                    "get_descriptions schema fetch failed",
-                    reason=r.error,
-                )
+        results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True, throw_exception=True)
         schemas = [r.data for r in results if r.data]
         rest = set()
         for s in schemas:
--- a/src/dremioai/api/dremio/search.py
+++ b/src/dremioai/api/dremio/search.py
@@ -261,30 +261,16 @@ async def get_search_results(
         paths = [p["path"] for p in data]
         skipped: List[Dict[str, Any]] = []
         if schemas := await get_schemas(paths, include_tags=True, flatten=True):
-            for ix, r in enumerate(schemas):
-                data[ix]["schema"] = r.data.get("schema") if r.data else None
-                if r.error:
-                    logger.info(
-                        "search_schema_fetch_failed",
-                        path=paths[ix],
-                        reason=r.error,
-                    )
-                    skipped.append({"path": paths[ix], "reason": r.error})
+            for ix, (schema, error) in enumerate(schemas):
+                if schema is None or error is not None:
+                    logger.error("Schema not found for search",
+                                 path=paths[ix], reason=error)
+                    schema = {}
+                data[ix]["schema"] = schema.get("schema")
 
-        # Drop rows for unresolvable paths and re-add them with `not_found=True`
-        # so the LLM can see the partial failure in the tool output.
-        data = [d for d, sr in zip(data, schemas) if not sr.error] if schemas else data
         df = pd.DataFrame(data=data)
-        if skipped:
-            df = pd.concat(
-                [
-                    df,
-                    pd.DataFrame(
-                        {"path": [s["path"] for s in skipped], "not_found": True}
-                    ),
-                ],
-                ignore_index=True,
-            )
+        # for any paths that we don't have schema, set not found
+        df["not_found"] = df.schema.isna()
         return df
 
     return EnterpriseSearchResultsWrapper(results=result)
--- a/src/dremioai/api/dremio/search.py
+++ b/src/dremioai/api/dremio/search.py
@@ -13,6 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
+import asyncio
 
 from pydantic import (
     BaseModel,
@@ -22,7 +23,6 @@ from pydantic import (
 )
 from typing import (
     Any,
-    Dict,
     List,
     Union,
     Optional,
@@ -32,7 +32,7 @@ from datetime import datetime
 from enum import auto
 from dremioai.config import settings
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
-from dremioai.api.dremio.catalog import get_schemas
+from dremioai.api.dremio.catalog import get_schemas, get_schema
 from dremioai import log
 import pandas as pd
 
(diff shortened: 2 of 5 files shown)


src/dremioai/api/dremio/catalog.py
```diff
@@ -191,10 +193,17 @@ async def get_schemas(
...
+        try:
+            return await get_schema(p, by_id, include_tags, flatten), None
+        except ClientResponseError as e:
+            return {}, f"HTTP {e.status}: {e.message}"
+        except Exception as e:
+            return {}, f"{type(e).__name__}: {e}"
 
     return await run_in_parallel(
-        [get_schema(p, by_id, include_tags, flatten) for p in dataset_path_or_ids]
+        [_safe_get_schema(p) for p in dataset_path_or_ids]
```
maxlepikhin: what is an example when search returns invalid path? That sounds like a search bug not MCP.

ssaumitra: @maxlepikhin When investigating DX-118395, I found that this specific issue occurs when customer's semantic layer contains VDS whose parent PDS is not found (and therefore schema extraction for that VDS fails). Check out my comment in DX-118395.

In this particular case, customer's semantic layer contains 20-30 valid datasets and 1-2 inconsistent VDS whose PDS is deleted. In my proposed solution, we will ignore those 1-2 inconsistent VDSs and MCP response show error to the user.

maxlepikhin: I see, so it is an issue in catalog - PDS was deleted w/o deletion of the views. @adam-christian-software @omarsmak 

Then, the defensive change is good.

maxlepikhin (dismissed):

src/dremioai/api/dremio/catalog.py
ssaumitra: In the Query engine space, I have observed that the catalog layer (be it Polaris, Nessie, Hive, etc) tends to keep VDS around, even if its parent PDS is deleted. We can ask catalog folks on the reason. 

The rationale could be that, the customer manually created a semantic layer with hundreds of VDSs that depend on each other. The customer intends to swap one PDS with another one and therefore chooses to delete it. However, if catalog chooses to delete all hundred VDSs that customer has manually created, it will result into bad customer experience. For that reason, catalog layer might have decided not to enforce consistency.

FWIW -- Query planning and execution layer also tend to be defensive in such cases. Before running SQL query, Planner always checks for dependent VDS to be present. And if they are not, planner throws an error. We can follow similar pattern in the other layers which use catalog directly.

src/dremioai/api/dremio/catalog.py
```diff
@@ -191,10 +193,17 @@ async def get_schemas(
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-) -> List[Dict[str, Any]]:
+) -> List[Tuple[Dict[str, Any], Optional[str]]]:
```
aniket-s-kulkarni: Add a 

```python
class GetSchemaResponse(BaseModel):
    schema: Optional[List[Tuple[Dict[str, Any]]
    error: Optional[str]]
```

Then change this signature to `-> GetSchemaResponse`

aniket-s-kulkarni (commented):
🟡 **Verdict: Must-fix design gaps** — 3 must fix, 1 should fix

**Must fix:**
- `get_schemas` return type should be a Pydantic `BaseModel` (e.g. `SchemaResult`) not a raw tuple — `src/dremioai/api/dremio/catalog.py`
- `get_descriptions` now silently swallows catalog errors, changing its original fail-fast contract — `src/dremioai/api/dremio/catalog.py`
- Missing server-side logging when a search hit's schema fetch fails — `src/dremioai/api/dremio/search.py`

**Should fix:**
- `get_lineage` still calls `get_schema` (singular) and accesses `response["id"]` directly — coordinate now to avoid a future silent break if `get_schema` is later changed to return `SchemaResult` — `src/dremioai/api/dremio/catalog.py:119`

---
*AI-assisted analysis*

src/dremioai/api/dremio/catalog.py
```diff
@@ -219,7 +228,9 @@ def extract_description(s: Dict[str, Any]) -> Dict[str, Any]:
     components = set()
     result = {}
     while True:
-        schemas = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
+        schemas = [
+            s for s, _ in await get_schemas(dataset_path_or_ids, by_id, include_tags=True) if s
```
aniket-s-kulkarni: **[MUST FIX]** This silently changes the error-handling contract of `get_descriptions`.

Before this PR, if any path failed inside `get_schemas`, the exception propagated through `asyncio.gather` and out of `get_descriptions` — callers got a hard failure. This PR silently drops the broken entries with no log and no signal.

`get_descriptions` is not the search tool. It should **not** be defensive. With the `SchemaResult` approach, preserve the original fail-fast behavior:

```python
results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
for r in results:
    if r.error:
        raise RuntimeError(f"schema fetch failed: {r.error}")
schemas = [r.data for r in results if r.data]
```

Only `get_search_results` / `SearchTableAndViews` should be defensive — other callers should continue to fail fast on catalog errors as they did before.

src/dremioai/api/dremio/catalog.py
```diff
@@ -191,10 +193,17 @@ async def get_schemas(
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-) -> List[Dict[str, Any]]:
+) -> List[Tuple[Dict[str, Any], Optional[str]]]:
+    async def _safe_get_schema(p):
```
aniket-s-kulkarni: **[MUST FIX]** Use a Pydantic `BaseModel` here instead of a raw tuple.

`List[Tuple[Dict[str, Any], Optional[str]]]` is untyped, undocumented, and inconsistent with the rest of this Pydantic-first codebase. Define a proper result model:

```python
class SchemaResult(BaseModel):
    data: Optional[Dict[str, Any]] = None
    error: Optional[str] = None
```

Return `List[SchemaResult]` from `get_schemas`. This gives callers named fields, IDE completion, and a stable shape that can be extended later (e.g. adding `status_code`) without another signature change.

src/dremioai/api/dremio/search.py
```diff
@@ -253,10 +253,16 @@ async def get_search_results(
...
+        skipped: List[Dict[str, Any]] = []
         if schemas := await get_schemas(paths, include_tags=True, flatten=True):
-            for ix, schema in enumerate(schemas):
-                data[ix]["schema"] = schema.get("schema")
-
-        return pd.DataFrame(data=data)
+            for ix, (schema, error) in enumerate(schemas):
+                data[ix]["schema"] = schema.get("schema") if schema else None
+                if error:
+                    skipped.append({"path": paths[ix], "reason": error})
```
aniket-s-kulkarni: **[MUST FIX]** Log the failure server-side before stashing it in `skipped`.

This is the exact scenario the fix targets — search returned a hit, but catalog couldn't find the schema (e.g. VDS whose parent PDS was deleted). Silently recording it in `skipped` surfaces the info to the LLM but emits nothing to operators watching logs during an incident.

Add a structured warning here (structlog is already used in this codebase):

```python
if error:
    logger.warning(
        "search_schema_fetch_failed",
        path=paths[ix],
        reason=error,
    )
    skipped.append({"path": paths[ix], "reason": error})
```

src/dremioai/api/dremio/catalog.py
ssaumitra: Done. Added `SchemaResult`.

src/dremioai/api/dremio/catalog.py
ssaumitra: @aniket-s-kulkarni AI might have hallucinated here. The root cause of DX-118395 was that the call to MCP server was failing when schema retrieval for one of the catalog entry failed.

I am therefore adding info logging but still letting MCP call to proceed when one of the VDS entry is corrupt. Can you please verify? I chose to add `info` logging and not error/warn, mainly because `get_schemas` failure is expected in the aforementioned cases. I do agree that we may want to keep an eye if we start gulping off legitimate problems.

src/dremioai/api/dremio/search.py
ssaumitra: Done. Added logging. IMO, `logger.info` is more appropriate because VDS being present with missing PDS, and thereby MCP not processing such VDS is an expected and as-designed behavior.

src/dremioai/api/dremio/catalog.py
adam-christian-software: Yes, the Catalog does not delete any View which are reliant on a Table which is deleted. That's not a feature of the Catalog.

src/dremioai/api/dremio/catalog.py
aniket-s-kulkarni: @ssaumitra not quite. 

> The root cause of DX-118395 was that the call to MCP server was failing when schema retrieval for one of the catalog entry failed.

The call to `SearchTableAndViews` tool was failing. It does a search under the covers and then does `get_schema` internally on each of the search matches. Your fix should apply only in this code path. It should not apply to `GetDescriptionOfTableOrSchema` path since that was an explicit call by LLM and it should see the error.

src/dremioai/api/dremio/catalog.py
```diff
@@ -219,7 +243,14 @@ def extract_description(s: Dict[str, Any]) -> Dict[str, Any]:
     components = set()
     result = {}
     while True:
-        schemas = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
+        results = await get_schemas(dataset_path_or_ids, by_id, include_tags=True)
+        for r in results:
+            if r.error is not None:
```
aniket-s-kulkarni: This is masking errors when called by `GetDescriptionOfTableOrSchema` and `GetSchemaOfTable` - that flow should continue to error out to the LLM

src/dremioai/api/dremio/search.py
```diff
@@ -253,10 +258,21 @@ async def get_search_results(
...
+                    logger.info(
+                        "search_schema_fetch_failed",
+                        path=paths[ix],
+                        reason=r.error,
+                    )
+                    skipped.append({"path": paths[ix], "reason": r.error})
+
+        df = pd.DataFrame(data=data)
+        if skipped:
+            df.attrs["skipped"] = skipped
```
aniket-s-kulkarni: @ssaumitra this is how the df looks like today 

```python
[ins] In [17]: asyncio.run(search.get_search_results(search="citibike", use_df=True))
Out[17]:

                                          path  ...                                             schema
0     [dremio_samples, nyc_citibike, citibike]  ...  {'tripduration': 'BIGINT', 'rideable_type': 'V...
1  [dremio_samples(1), nyc_citibike, citibike]  ...  {'tripduration': 'BIGINT', 'rideable_type': 'V...

[2 rows x 6 columns]

[ins] In [22]: df.schema[0]
Out[22]:

{
    'tripduration': 'BIGINT',
    'rideable_type': 'VARCHAR',
    'starttime': 'TIMESTAMP WITH PRECISION',
    'stoptime': 'TIMESTAMP WITH PRECISION',
    'start station id': 'BIGINT',
    'start station name': 'VARCHAR',
    'start station latitude': 'DOUBLE',
    'start station longitude': 'DOUBLE',
    'end station id': 'BIGINT',
    'end station name': 'VARCHAR',
    'end station latitude': 'DOUBLE',
    'end station longitude': 'DOUBLE',
    'usertype': 'VARCHAR'
}
```

Where as what we want is - notice the 'not found' column

```path
Out[27]:

                                          path                                           name  ...                                             schema not found
0     [dremio_samples, nyc_citibike, citibike]     "dremio_samples"."nyc_citibike"."citibike"  ...  {'tripduration': 'BIGINT', 'rideable_type': 'V...       NaN
1  [dremio_samples(1), nyc_citibike, citibike]  "dremio_samples(1)"."nyc_citibike"."citibike"  ...  {'tripduration': 'BIGINT', 'rideable_type': 'V...       NaN
0                              [unknown, path]                                            NaN  ...                                                NaN      True
```


so add 

```python
if skipped:
    df = pd.concat([df, pd.DataFrame({
        "path": [s["path"] for s in skipped],
        "not_found": True}])
```

Do a similar change to `EnterpriseSearchResultsWrapper` - add a `not_found` attribute to it.

src/dremioai/tools/tools.py
```diff
@@ -533,8 +533,17 @@ async def invoke(self, query: str) -> Dict[str, Any]:
                 for category in (search.Category.TABLE, search.Category.VIEW)
             ]
         )
-        res = pd.concat(res)
-        return {"results": res.to_dict(orient="records")}
+        skipped = [s for df in res for s in df.attrs.get("skipped", [])]
```
aniket-s-kulkarni: This is no longer necessary.. the json will have not_found

src/dremioai/api/dremio/catalog.py
ssaumitra: Valid point. I missed that `get_schemas` is also used in `GetDescriptionOfTableOrSchema` path. I will correct this now.

src/dremioai/api/dremio/catalog.py
ssaumitra: Done. This is no longer masking any errors from `GetDescriptionOfTableOrSchema` and `GetSchemaOfTable`.

src/dremioai/tools/tools.py
ssaumitra: Done. Removed now. Adding `not_found ` in a follow-up commit.

src/dremioai/api/dremio/catalog.py
ssaumitra: Done

src/dremioai/api/dremio/search.py
ssaumitra: Done. For 
> Do a similar change to EnterpriseSearchResultsWrapper - add a not_found attribute to it.

I added not_found attribute. But please note that when `use_df` is false, we do not call `get_schemas` . So the schema not found error is not hit in that codepath.

ssaumitra:
@aniket-s-kulkarni I have incorporated all of your review responses. Can you please review again?

src/dremioai/api/dremio/catalog.py
```diff
@@ -186,15 +190,34 @@ async def _get(id, suffix):
...
+
+
 async def get_schemas(
     dataset_path_or_ids: List[Union[List[str], str]],
     by_id: Optional[bool] = False,
     include_tags: Optional[bool] = False,
     flatten: Optional[bool] = False,
-) -> List[Dict[str, Any]]:
+    throw_exception: Optional[bool] = False,
+) -> List[SchemaResult]:
```
aniket-s-kulkarni: it should be `SchemaResult` not `List[SchemaResult]`

src/dremioai/api/dremio/search.py
```diff
@@ -253,10 +259,32 @@ async def get_search_results(
...
-        return pd.DataFrame(data=data)
+            for ix, r in enumerate(schemas):
+                data[ix]["schema"] = r.data.get("schema") if r.data else None
+                if r.error:
+                    logger.info(
+                        "search_schema_fetch_failed",
+                        path=paths[ix],
+                        reason=r.error,
+                    )
+                    skipped.append({"path": paths[ix], "reason": r.error})
```
aniket-s-kulkarni: I am going to suggest a simplification here to prevent filtering out and then adding it back in. 


```python
for ix, (schema, error) in enumerate(schemas):
    if schema is None or error is not None
        logger.error("Schema not found for search",
           path=paths[ix], reason=error)
        schema = {}
    data[ix]["schema"] = schema.get("schema")

df = pd.DataFrame(data=data)
# for any paths that we don't have schema, set not found
df["not_found"] = df.schema.isna()
```

and now remove rest of the code.

src/dremioai/api/dremio/catalog.py
ssaumitra: Done

src/dremioai/api/dremio/search.py
ssaumitra: Done

aniket-s-kulkarni (dismissed):

src/dremioai/api/dremio/search.py
```diff
@@ -253,10 +259,19 @@ async def get_search_results(
...
-        return pd.DataFrame(data=data)
+            for ix, (schema, error) in enumerate(schemas):
+                if schema is None or error is not None:
+                    logger.error("Schema not found for search",
+                                 path=paths[ix], reason=error)
+                    schema = {}
+                if "schema" in schema:
+                    data[ix]["schema"] = schema.get("schema")
+                else:
+                    data[ix]["not_found"] = True
```
aniket-s-kulkarni: @ssaumitra looking good; sorry just one more 

```python
for ix, (schema, error) in enumerate(schemas):
    if schema is None or error is not None:
        logger.error("Schema not found for search",
            path=paths[ix], reason=error)
    
    data[ix]["schema"] = schema.get("schema")
 
df = pd.DataFrame(data=data)
df["not_found"] = df.schema.isna()
```

src/dremioai/api/dremio/search.py
```diff
@@ -253,10 +259,19 @@ async def get_search_results(
 
         data = [r.catalog.as_df_dict() for r in result]
         paths = [p["path"] for p in data]
+        skipped: List[Dict[str, Any]] = []
         if schemas := await get_schemas(paths, include_tags=True, flatten=True):
-            for ix, schema in enumerate(schemas):
-                data[ix]["schema"] = schema.get("schema")
-
-        return pd.DataFrame(data=data)
+            for ix, (schema, error) in enumerate(schemas):
```
aniket-s-kulkarni: @ssaumitra looking good; sorry but for returning `EnterpriseSearchResultsWrapper` we need to make one more small change


```python
not_found = []
if schemas := ...
    for ix, (schema, error) in enumerate(schemas):
        if schema is None or error is not None:
            logger.error("Schema not found for search",
                path=paths[ix], reason=error)
            not_found.append(paths[ix])
        data[ix]["schema"] = schema.get("schema")
 
    df = pd.DataFrame(data=data)
    df["not_found"] = df.schema.isna()
    return df

return EnterpriseSearchResultsWrapper(
    result = [r for r in result if r.catalog.path not in not_found]
    not_found = not_found)
```

ssaumitra: > return EnterpriseSearchResultsWrapper(
    result = [r for r in result if r.catalog.path not in not_found]
    not_found = not_found)

When `use_df` is False, we use [this codepath](https://github.com/dremio/dremio-mcp/blob/1335b7b8088f10bd6a47202849cb2b54bd4daad4/src/dremioai/api/dremio/search.py#L229-L255) where `get_schemas` is not called and `not_found` are not initialized. Should we change that and start calling `get_schemas` over there as well?

> `df["not_found"] = df.schema.isna() `

@aniket-s-kulkarni I made this change in the previous iteration that caused the test to fail 

Basically the test failure was arising from 
```
df.schema.isna()
```
When `df` is empty, `df.schema` throws exceptions. 
I changed the code slightly in order to make it agnostic to `df` contents (and therefore the test passes now). Basically I am preparing a dict with all fields in the form that we would like to return and then constructing dataframe.

aniket-s-kulkarni: > > return EnterpriseSearchResultsWrapper(
> > result = [r for r in result if r.catalog.path not in not_found]
> > not_found = not_found)
> 
> When `use_df` is False, we use [this codepath](https://github.com/dremio/dremio-mcp/blob/1335b7b8088f10bd6a47202849cb2b54bd4daad4/src/dremioai/api/dremio/search.py#L229-L255) where `get_schemas` is not called and `not_found` are not initialized. Should we change that and start calling `get_schemas` over there as well?

@ssaumitra yes good point, thats exactly what I am suggesting.


> 
> > `df["not_found"] = df.schema.isna() `
> 
> @aniket-s-kulkarni I made this change in the previous iteration that caused the test to fail
> 
> Basically the test failure was arising from
> 
> ```
> df.schema.isna()
> ```
> 
> When `df` is empty, `df.schema` throws exceptions. I changed the code slightly in order to make it agnostic to `df` contents (and therefore the test passes now). Basically I am preparing a dict with all fields in the form that we would like to return and then constructing dataframe.

@ssaumitra then change it to 

```python
if not df.empty:
    df["not_found"] = df.schema.isna()
```

ssaumitra: >> Should we change that and start calling get_schemas over there as well?
>
> yes good point, thats exactly what I am suggesting.

@aniket-s-kulkarni In the current codepath, I see that `EnterpriseSearchResultsWrapper` is returning instances of `EnterpriseSearchResultsObject` directly. If we want to do what you are proposing, the cleanest way is to add additional field in `EnterpriseSearchCatalogObject`. I am moving the `get_schema` and relevant code to the class function of `EnterpriseSearchCatalogObject`.

ssaumitra: In that case, we won't need `not_found` field in `EnterpriseSearchResultsWrapper` that you advised to add earlier. I am changing those things now.

ssaumitra: @aniket-s-kulkarni After I moved schema population logic into `EnterpriseSearchCatalogObject` I realized that changes in catalog.py in `get_schemas` function as well as the cascading changes (like `SchemaResult` class, etc) are no longer needed. The fix is now entirely in `search.py`. Please review.

In case of any follow-up, let's huddle and sort the review feedbacks out.

src/dremioai/api/dremio/search.py
```diff
@@ -155,6 +160,22 @@ class EnterpriseSearchCatalogObject(BaseModel):
     modified_at: Optional[datetime] = Field(default=None, alias="lastModifiedAt")
     func_sql: Optional[str] = Field(default=None, alias="functionSql")
     owner: Optional[EnterpriseSearchUserOrRoleObject] = None
+    schema: Optional[Any] = None
+    schema_not_found: Optional[bool] = None
```
aniket-s-kulkarni:
```suggestion
    schema: Optional[Dict[str, Any]] = None
```

src/dremioai/api/dremio/search.py
```diff
@@ -155,6 +160,22 @@ class EnterpriseSearchCatalogObject(BaseModel):
...
+                data = await get_schema(dataset_path_or_id=self.path, include_tags=True, flatten=True)
+            except Exception as e:
+                logger.error("Schema not found for search",
+                             path=self.path, reason=f"{type(e).__name__}: {e}")
+                data = {}
+            if "schema" in data:
+                self.schema = data.get("schema")
+                self.schema_not_found = False
+            else:
+                self.schema_not_found = True
```
aniket-s-kulkarni:
```suggestion
        if self.path:
            try:
                data = await get_schema(dataset_path_or_id=self.path, include_tags=True, flatten=True)
                self.schema = data.get('schema') if data else None
            except Exception as e:
                logger.error("Schema not found for search",
                             path=self.path, reason=str(e))
```

src/dremioai/api/dremio/search.py
```diff
@@ -163,6 +184,8 @@ def as_df_dict(self):
             "type": self.type,
             "tags": ",".join(self.labels),
             "description": self.wiki,
+            "schema": self.schema,
+            "not_found": self.schema_not_found,
```
aniket-s-kulkarni:
```suggestion
            "schema": self.schema,
```

src/dremioai/api/dremio/search.py
```diff
@@ -247,16 +270,12 @@ async def get_search_results(
...
-
         data = [r.catalog.as_df_dict() for r in result]
-        paths = [p["path"] for p in data]
-        if schemas := await get_schemas(paths, include_tags=True, flatten=True):
-            for ix, schema in enumerate(schemas):
-                data[ix]["schema"] = schema.get("schema")
-
-        return pd.DataFrame(data=data)
-
+        df = pd.DataFrame(data=data)
```
aniket-s-kulkarni, line 272:
```suggestion
        df = pd.DataFrame(data=data)
        if not df.empty:
            df["not_found"] = df.schema.isna()
```

ssaumitra: @aniket-s-kulkarni With the changes that you are proposing, `not_found` information will be removed from `EnterpriseSearchResultsWrapper`. This contradicts with your own ask in the previous comment
Screenshot 2026-04-25 at 1 43 23 AM


Let's have a huddle to sort this out.

aniket-s-kulkarni (approved):

PR #106, opened 2026-04-29
DX-119753: Switch Streamable HTTP to stateless mode + add session-aware diagnostics

aniket-s-kulkarni:
## Summary

Jira:
- https://dremio.atlassian.net/browse/DX-119753
- https://dremio.atlassian.net/browse/DX-118422

Adds MCP request correlation and transport diagnostics for Streamable HTTP, then switches Streamable HTTP to stateless mode to avoid stale in-memory MCP session failures after server/pod restarts.

## RCA

While testing Claude through ngrok against a local MCP server, the server was restarted after Claude had already initialized an MCP Streamable HTTP session. Claude then reused the old session id on the next tool call:

```http
POST /mcp/<project_id> HTTP/1.1
Accept: application/json, text/event-stream
Mcp-Protocol-Version: 2025-06-18
Mcp-Session-Id: cc09fe87d1944e0996d15687c69286de

{"method":"tools/call","params":{"name":"RunSqlQuery","arguments":{"query":"SELECT 1"}},"jsonrpc":"2.0","id":1}
```

The restarted server returned:

```http
HTTP/1.1 400 Bad Request

Bad Request: No valid session ID provided
```

The request had a valid protocol version and did include `Mcp-Session-Id`. The failure was that the Python MCP SDK stores Streamable HTTP session transports in process memory. After a restart, the server's `_server_instances` map is empty, so Claude's old session id is unknown.

The MCP Streamable HTTP spec expects stale/expired session recovery via `404`, allowing clients to reinitialize. The current Python SDK returns `400` for this unknown-session path, which can prevent Claude from recovering cleanly.

## Changes

- Adds `structlog.contextvars.merge_contextvars` to the logging pipeline.
- Binds MCP request context for the entire tool invocation scope with `bound_contextvars`, so logs emitted below the tool wrapper include the same correlation fields.
- Adds a per-call `tool_invocation_id` so multiple calls in the same MCP session can be distinguished.
- Adds `MCPTransportLoggingMiddleware` around `/mcp...` requests to log safe 3xx and above transport diagnostics before a tool invocation exists.
- Enables FastMCP `stateless_http=True` for Streamable HTTP so server restarts and non-sticky routing do not invalidate idle clients via process-local MCP session state.

Logged tool invocation fields include:
- tool name, outcome, duration
- MCP session id and protocol version when available
- JSON-RPC request id
- tool invocation id
- client host, request method/path
- project id and authenticated user id when available
- error type/message for failed invocations

Logged transport failure fields include:
- status code
- response error body, truncated to 2048 bytes
- request/response MCP session id when present
- MCP protocol version
- request method/path/client
- `Accept` and `Content-Type`
- project id when available

This intentionally does not log bearer tokens, request bodies, tool arguments, SQL text, result data, or Authorization headers. It also avoids adding session/user labels to Prometheus metrics to prevent high-cardinality metrics.

## Repro

1. Use a config like:

```yaml
dremio:
  enable_search: true
  allow_dml: true
  project_id: DREMIO_DYNAMIC
  uri: https://api.dev.dremio.site
tools:
  server_mode: FOR_DATA_PATTERNS
```

2. Start the server:

```bash
uv run dremio-mcp-server run --enable-streaming-http --no-log-to-file --log-level DEBUG
```

3. Expose it:

```bash
ngrok http 8000
```

4. Add the ngrok project-scoped URL to Claude as an MCP connector:

```text
https://<ngrok-host>/mcp/<project_id>
```

5. Call `RunSqlQuery` with `SELECT 1`.
6. Restart the MCP server while leaving the Claude connector/session in place.
7. Call `RunSqlQuery` again.

Pre-fix, Claude reuses the stale `Mcp-Session-Id` and the SDK returns `400 Bad Request: No valid session ID provided`.

## Testing

- `uv run pytest tests/servers/test_jwks_verifier.py::TestMCPTransportLoggingMiddleware tests/servers/test_jwks_verifier.py::TestStreamableHttpInit tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run pytest tests/test_log.py -q`
- `uv run pytest tests/config/test_launchdarkly_integration.py::test_log_level_refresh_updates_level tests/config/test_launchdarkly_integration.py::test_log_level_refresh_no_change_when_same tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run black --check --diff --workers 1 src/dremioai/servers/mcp.py`

### DIFF
--- a/src/dremioai/log.py
+++ b/src/dremioai/log.py
@@ -105,6 +105,7 @@ def configure(enable_json_logging=None, to_file=False):
         else structlog.dev.ConsoleRenderer()
     )
     processors = [
+        structlog.contextvars.merge_contextvars,
         structlog.processors.add_log_level,
         structlog.processors.TimeStamper(),
         structlog.stdlib.filter_by_level,
--- a/src/dremioai/servers/mcp.py
+++ b/src/dremioai/servers/mcp.py
@@ -59,6 +59,7 @@ from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoin
 from starlette.requests import Request
 from starlette.responses import Response
 from starlette.responses import Response as StarletteResponse
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 from structlog.contextvars import bound_contextvars
 from typer import Argument, BadParameter, Option, Typer
 from yaml import dump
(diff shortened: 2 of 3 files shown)


ssaumitra (dismissed):
lgtm

aniket-s-kulkarni:
Implemented MCP tool invocation log-context propagation in `aae60bc`.

What changed:
- `dremioai.log.configure()` now includes `structlog.contextvars.merge_contextvars`, so context bound with structlog contextvars is merged into every `dremioai.log` event.
- `make_logged_invoke()` now binds the MCP request context for the whole tool invocation scope using `bound_contextvars(...)` instead of attaching the session fields only to the wrapper's success/error log.
- The bound context includes `mcp_session_id`, JSON-RPC request id, protocol version, client, method/path, project/user ids when available, `tool`, and a per-call `tool_invocation_id`.
- The context is automatically reset when the invocation exits, so it does not leak into later logs.

Added regression coverage:
- `TestMakeLoggedInvoke::test_propagates_mcp_request_context_to_nested_logs` emits a log from inside the wrapped tool function and verifies it contains the MCP session/request/tool context.
- The same test emits another log after the invocation and verifies the session context is no longer present.

Local verification:
- `uv run pytest tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run pytest tests/test_log.py -q`
- `uv run pytest tests/config/test_launchdarkly_integration.py::test_log_level_refresh_updates_level tests/config/test_launchdarkly_integration.py::test_log_level_refresh_no_change_when_same tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run black --check --diff src/dremioai/log.py src/dremioai/servers/mcp.py`

ssaumitra (dismissed):

aniket-s-kulkarni:
Added follow-up changes in `78914d5` for the Claude/ngrok restart troubleshooting case.

What changed:
- Streamable HTTP now initializes FastMCP with `stateless_http=True`, so the server no longer depends on in-memory MCP session ids for subsequent HTTP requests. This avoids idle clients breaking when the local server/pod restarts or when requests land on a different process.
- Added `MCPTransportLoggingMiddleware` around the Streamable HTTP app. For `/mcp...` responses with status >= 400 it logs safe transport diagnostics:
  - HTTP method/path/client/status
  - request `mcp-session-id`
  - request `mcp-protocol-version`
  - request `Accept` and `Content-Type`
  - response `mcp-session-id`, if any
  - response error body, truncated to 2048 bytes
  - project id from the project path middleware when available
- The middleware does not log Authorization headers or request bodies/tool arguments.

Added tests:
- transport 400 logging includes session/protocol/error body context and excludes bearer token contents
- streamable HTTP init sets `stateless_http=True`
- existing tool-level context propagation tests still pass

Local verification:
- `uv run pytest tests/servers/test_jwks_verifier.py::TestMCPTransportLoggingMiddleware tests/servers/test_jwks_verifier.py::TestStreamableHttpInit tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run pytest tests/test_log.py -q`
- `uv run pytest tests/config/test_launchdarkly_integration.py::test_log_level_refresh_updates_level tests/config/test_launchdarkly_integration.py::test_log_level_refresh_no_change_when_same tests/servers/test_jwks_verifier.py::TestMakeLoggedInvoke -q`
- `uv run black --check --diff --workers 1 src/dremioai/servers/mcp.py`

ssaumitra (approved):
As discussed in the call, we found the root cause for HTTP 400/401 errors while adding the logging and we are fixing in the same PR.

Please adjust thePR title and description and add information about `stateless_http` changes.

Otherwise the changes look good to me.

PR #108, opened 2026-05-07
DX-120297: updating the MCP to verify ES256 signatures

ajulka-drem:
Summary

  - Supporting ES256 signed tokens (ECDSA); the MCP JWKS verifier previously only accepted RS256
  - Updated algorithms list in JWKSVerifier.verify() to accept RS256, ES256, ES384, and ES512, covering both legacy RSA and all ECDSA variants.
  - Added tests/servers/test_jwks_integration.py — a skipped-by-default integration test harness that validates a real JWKS endpoint and JWT, including
  tampered-token and garbage-token cases.

  Test plan

  - Fill in JWKS_URL and TOKEN in tests/servers/test_jwks_integration.py with a live ES256-signed token and run uv run pytest
  tests/servers/test_jwks_integration.py -v -s to confirm the verifier accepts the token and returns valid claims.
  - Confirm a tampered token returns None (test already covers this).
  - Confirm existing unit tests still pass: uv run pytest tests/.

### DIFF
--- a/src/dremioai/servers/jwks_verifier.py
+++ b/src/dremioai/servers/jwks_verifier.py
@@ -126,7 +126,7 @@ class JWKSVerifier:
         claims = pyjwt.decode(
             token,
             signing_key.key,
-            algorithms=["RS256"],
+            algorithms=["RS256", "ES256", "ES384", "ES512"],
             options={
                 "verify_aud": False,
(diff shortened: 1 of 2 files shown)


ssaumitra (approved):
lgtm

PR #113, opened 2026-05-15
DX-120975: Fix endpoint

AfonsoP1nto:
Update AI tool API calls to use the same project-aware endpoint logic as the rest of the Dremio client.

`list_tools()` and `invoke_tool()` now call `/v0/projects/{project_id}/ai/tools` when a project ID is configured.
When no project ID is present, both paths fall back to /api/v3/ai/tools.
Updated tests to cover both project-scoped and non-project-scoped requests, and removed the hardcoded `/api/v4/ai/tools` Software-only paths.

### Validation

`python3 -m py_compile src/dremioai/api/dremio/ai_tools.py tests/api/dremio/test_ai_tools.py`

Also added a fix to the `test_stdlib_logs_are_rendered_as_json` test, which had a pre-existing error. The e2e tests that start real uvicorn servers trigger uvicorn's configure_logging(), which applies its default LOGGING_CONFIG — this explicitly sets uvicorn.access.propagate = False. That setting persisted across tests, so when test_stdlib_logs_are_rendered_as_json ran later, the uvicorn.access logger silently swallowed the log message instead of propagating it to the root handler.

Added a reset_uvicorn_logger_propagation autouse fixture to conftest.py that resets propagate=True and clears handlers on uvicorn loggers after each test.

### DIFF
--- a/src/dremioai/api/dremio/ai_tools.py
+++ b/src/dremioai/api/dremio/ai_tools.py
@@ -20,6 +20,7 @@ from urllib.parse import quote
 
 from aiohttp import ClientResponseError
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
+from dremioai.config import settings
 from dremioai.log import logger
 
 log = logger(__name__)
--- a/src/dremioai/api/dremio/ai_tools.py
+++ b/src/dremioai/api/dremio/ai_tools.py
@@ -64,10 +64,9 @@ class InvokeToolResponse(BaseModel):
 async def list_tools() -> ListToolsResponse:
     try:
         client = AsyncHttpClient()
-        return await client.get(
-            "/api/v4/ai/tools",
-            deser=ListToolsResponse,
-        )
+        project_id = settings.instance().dremio.project_id
+        endpoint = f"/v0/projects/{project_id}" if project_id else "/api/v3"
+        return await client.get(f"{endpoint}/ai/tools", deser=ListToolsResponse)
     except ClientResponseError as e:
         log.exception("Failed to list AI tools")
         return ListToolsResponse(error=f"HTTP {e.status} {e.message}")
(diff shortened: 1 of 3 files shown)


src/dremioai/api/dremio/ai_tools.py
```diff
@@ -64,10 +65,9 @@ def is_empty(self) -> bool:
 async def list_tools() -> ListToolsResponse:
     try:
         client = AsyncHttpClient()
-        return await client.get(
-            "/api/v4/ai/tools",
-            deser=ListToolsResponse,
-        )
+        project_id = settings.instance().dremio.project_id
+        endpoint = f"/v0/projects/{project_id}" if project_id else "/api/v3"
```
aniket-s-kulkarni: `/api/v4` --> maps to --> `/v1/projects/{project_id}/..` I would recommend trying out the dremio url by hand (curl) for a quick test

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -80,8 +80,10 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
     safe_name = quote(tool_name, safe="")
     try:
         client = AsyncHttpClient()
+        project_id = settings.instance().dremio.project_id
+        endpoint = f"/v0/projects/{project_id}" if project_id else "/api/v3"
```
aniket-s-kulkarni: ditto.

aniket-s-kulkarni: for software it will be `/api/v4` not v3

src/dremioai/api/dremio/ai_tools.py
aniket-s-kulkarni: for software it would v4 not v3

tests/api/dremio/test_ai_tools.py
```diff
@@ -42,7 +51,7 @@ async def test_list_tools_returns_tools(mock_settings_instance):
 @pytest.mark.asyncio
 async def test_list_tools_returns_input_schema(mock_settings_instance):
     with HttpMockFramework() as mock:
-        mock.load_mock_data(r"/api/v4/ai/tools$", "ai_tools/list_tools.json")
+        mock.load_mock_data(r"/v0/projects/[^/]+/ai/tools$", "ai_tools/list_tools.json")
```
aniket-s-kulkarni: ditto.

aniket-s-kulkarni (changes requested):

src/dremioai/api/dremio/ai_tools.py
AfonsoP1nto: Done, I'm spinning up a cluster to test changes, these changes should be good and the following changes will be tested afterwards

aniket-s-kulkarni (approved):

PR #114, opened 2026-05-27
DX-119494 Add protected resource metadata for Langdock OAuth reauth

aniket-s-kulkarni:
## Summary
- add RFC 9728 protected-resource metadata for project-scoped MCP endpoints
- advertise `resource_metadata` on 401 Bearer challenges and reuse shared metadata helpers
- cover real and mock auth flows with e2e/mock tests and document the DX-119494 Langdock reauth failure mode

## Context
DX-119494 tracks Langdock custom OAuth integrations timing out after the access token aged out, forcing manual reauth instead of a clean refresh/reauth flow. The customer-configured MCP endpoint includes `/mcp/<project_id>`, so OAuth clients need the path-inserted protected-resource metadata URL (`/.well-known/oauth-protected-resource/mcp/<project_id>`) to discover the correct reauth target.

## Testing
- `uv run pytest tests/e2e/test_mcp_e2e.py -q -k "oauth_metadata_includes_registration_endpoint or protected_resource_metadata_and_401_challenge"`
- `uv run pytest tests/servers/test_mock_auth.py tests/e2e/test_mcp_e2e.py -q -k "oauth or protected_resource_metadata"`

### DIFF
--- a/src/dremioai/servers/mcp.py
+++ b/src/dremioai/servers/mcp.py
@@ -68,7 +68,10 @@ from yaml import dump
 
 from dremioai import log
 from dremioai.api.oauth2 import get_oauth2_tokens
-from dremioai.api.oauth_metadata import OAuthMetadataRFC8414
+from dremioai.api.oauth_metadata import (
+    OAuthMetadataRFC8414,
+    OAuthProtectedResourceMetadata,
+)
 from dremioai.config import settings
 from dremioai.config.feature_flags import FeatureFlagManager
 from dremioai.metrics.registry import get_metrics_app
@@ -196,11 +199,27 @@ class RequireAuthWithWWWAuthenticateMiddleware(BaseHTTPMiddleware):
                 project_id=ProjectIdMiddleware.get_project_id(),
                 endpoint=endpoint,
             )
+            # This middleware is the last step that emits the HTTP 401 after the
+            # auth backend has already classified the request as unauthenticated.
+            # RFC 9728 Section 5.1 defines the `resource_metadata` auth-param on
+            # the Bearer challenge so clients can discover the protected resource
+            # metadata document after that 401. If a Bearer token was supplied,
+            # RFC 6750 Section 3.1 allows `error="invalid_token"`, which lets
+            # OAuth-aware MCP clients distinguish "missing token" from "token
+            # rejected upstream by verification/authentication".
+            has_bearer_token = request.headers.get("authorization", "").startswith(
+                "Bearer "
+            )
+            www_authenticate = (
+                f'Bearer resource_metadata="{build_resource_metadata_url(request)}"'
+            )
+            if has_bearer_token:
+                www_authenticate += ', error="invalid_token"'
             # Return 401 with WWW-Authenticate header
             return StarletteResponse(
                 content="Unauthorized",
                 status_code=401,
-                headers={"WWW-Authenticate": "Bearer"},
+                headers={"WWW-Authenticate": www_authenticate},
             )
 
         # User is authenticated, proceed with the request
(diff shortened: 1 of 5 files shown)


ssaumitra (commented):
🔴 **Verdict: Issues found** — 1 must fix, 1 should fix

**Inspect first:**
- `src/dremioai/servers/mcp.py` — new OAuth discovery URL is built from request forwarding headers on an unauthenticated path.
- `tests/e2e/test_mcp_e2e.py` — coverage exercises the root MCP path but not the project-scoped path that motivated the change.

🔴 **Must fix:**
- [Security] Avoid reflecting untrusted forwarded headers into the OAuth `resource_metadata` discovery URL (`src/dremioai/servers/mcp.py:233`).

🟡 **Should fix:**
- [Tests] Add project-scoped `/mcp/{project_id}` coverage for protected-resource metadata and the 401 challenge (`tests/e2e/test_mcp_e2e.py:122`).

---
*Generated by AI-assisted analysis.*

tests/e2e/test_mcp_e2e.py
```diff
@@ -110,6 +118,29 @@ async def test_oauth_metadata_includes_registration_endpoint(
             ), f"registration_endpoint should end with /oauth/register, got: {data['registration_endpoint']}"
 
 
+@pytest.mark.asyncio
+async def test_protected_resource_metadata_and_401_challenge(
+    mock_config_dir, logging_server, logging_level
+):
+    async with http_streamable_mcp_server(logging_server, logging_level) as sf:
+        metadata_url = _protected_resource_metadata_url(sf.mcp_server.url)
```
ssaumitra, line 136: 🟡 **SHOULD FIX** · Tests · High confidence

**Exercise the project-scoped MCP path**

This test validates the root `/mcp/` URL, but DX-119494 is specifically about customer-configured `/mcp/{project_id}` endpoints. A regression in the path-inserted protected-resource metadata URL or the project-scoped 401 challenge would still pass here.

Please add or parameterize coverage with `http_streamable_mcp_server(..., project_id=<id>)` and assert the protected-resource document plus `WWW-Authenticate` challenge for `/.well-known/oauth-protected-resource/mcp/{project_id}`.

src/dremioai/servers/mcp.py
```diff
@@ -212,6 +230,62 @@ class Transports(StrEnum):
     streamable_http = "streamable-http"
 
 
+def request_base_url(request: Request) -> str:
+    headers = getattr(request, "headers", {}) or {}
+    url = getattr(request, "url", None)
+    scheme = headers.get(
+        "x-forwarded-proto", getattr(url, "scheme", None) or "http"
```
ssaumitra: 🔴 **MUST FIX** · Security · High confidence

**Do not reflect untrusted forwarding headers into OAuth discovery**

`request_base_url()` trusts `X-Forwarded-Proto` / `X-Forwarded-Host`, and the result is reflected into the unauthenticated `WWW-Authenticate: Bearer resource_metadata="..."` challenge. If those headers are client-controlled or not sanitized by ingress, an attacker can poison the OAuth discovery URL advertised to MCP clients.

Please derive this from a configured canonical origin, or only honor forwarded headers behind explicit trusted-proxy handling. The auth-param value should also be validated/escaped before it is emitted.

tests/e2e/test_mcp_e2e.py
aniket-s-kulkarni: Agreed, and I expanded the test to cover both the root `/mcp/` path and project-scoped `/mcp/{project_id}` paths. That new coverage also exposed a real path extraction bug in the protected-resource handler, which is fixed in the latest push.

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Good catch — updated this so the unauthenticated `resource_metadata` challenge no longer trusts raw `X-Forwarded-*` headers. It now uses a configured canonical override when present and otherwise falls back to the observed request URL.

aniket-s-kulkarni:
> 🔴 **Verdict: Issues found** — 1 must fix, 1 should fix
>
> **Inspect first:**
> - `src/dremioai/servers/mcp.py` — new OAuth discovery URL is built from request forwarding headers on an unauthenticated path.
> - `tests/e2e/test_mcp_e2e.py` — coverage exercises the root MCP path but not the project-scoped path that motivated the change.
>
> 🔴 **Must fix:**
> - [Security] Avoid reflecting untrusted forwarded headers into the OAuth `resource_metadata` discovery URL (`src/dremioai/servers/mcp.py:233`).
>
> 🟡 **Should fix:**
> - [Tests] Add project-scoped `/mcp/{project_id}` coverage for protected-resource metadata and the 401 challenge (`tests/e2e/test_mcp_e2e.py:122`).
>
> ---
>*Generated by AI-assisted analysis.*

**📬 Review comments addressed** — 2026-05-27

| Finding | Resolution | Notes |
|---------|------------|-------|
| [Security] Avoid reflecting untrusted forwarded headers into the OAuth `resource_metadata` discovery URL (`src/dremioai/servers/mcp.py:233`) | ✅ Implemented | Stopped trusting raw `X-Forwarded-*` values in the unauthenticated challenge path; the code now uses a configured canonical override when present and otherwise falls back to the observed request URL. |
| [Tests] Add project-scoped `/mcp/{project_id}` coverage for protected-resource metadata and the 401 challenge (`tests/e2e/test_mcp_e2e.py:122`) | ✅ Implemented | Expanded coverage to root and project-scoped MCP endpoints; the new test also exposed and fixed project path extraction in the protected-resource handler. |

ssaumitra:
Code changes look good to me. `tests/config/test_launchdarkly_integration.py::test_flag_keys_match_golden` needs to be fixed though. Can you please fix that? I can then approve right-away.

ssaumitra (dismissed):
LGTM

ssaumitra (approved):

PR #115, opened 2026-05-27
DX-121841: Upgrade MCP Python SDK from 1.10.0 to 1.27.1 (Phase 1)

ssaumitra:
## Summary

This PR upgrades the dremio-mcp server's MCP Python SDK from 1.10.0 to 1.26.0+ (resolved to **1.27.1**, the latest stable v1), with comprehensive compatibility fixes for breaking changes in Starlette 1.0 and SDK 1.14+.

### Dependency Updates

- `mcp`: `1.10.0` → `>=1.26.0,<2` (resolved to 1.27.1). The `<2` upper bound guards against the v2 pre-release which breaks all `mcp.server.fastmcp.*` imports.
- `starlette`: `>=0.49.1` → `>=1.0.0` (uv.lock resolves to 1.0.0)
- `pydantic`: `>=2.10.6` → `>=2.11.0`

### Architecture & Breaking Changes

**ProjectIdMiddleware: BaseHTTPMiddleware → Raw ASGI Middleware**
Starlette 1.0's `BaseHTTPMiddleware.call_next()` captures the original ASGI scope via closure, meaning any path rewrites applied in `dispatch()` are silently ignored by the middleware's wrapped app. This caused the SDK's stricter `/mcp`-only route matching (added in 1.14+) to reject `/mcp/{project_id}/` paths with 404 errors. Converted to a raw ASGI middleware that modifies `scope["path"]` and `scope["raw_path"]` in-place before forwarding to the app.

**DNS Rebinding Protection: opt-in disable via `--disable-dns-rebinding-protection`**
SDK 1.14+ auto-enables DNS rebinding protection in FastMCP when bound to 127.0.0.1/localhost/::1, restricting accepted Host headers to those exact values. This breaks deployments behind a reverse proxy or tunnel where the public-facing hostname differs from the bind interface—incoming requests are rejected with HTTP 421 "Invalid Host header" before reaching the auth layer.

Rather than unconditionally disabling the protection, a new `--disable-dns-rebinding-protection` CLI flag is added to the `run` command. By default the SDK's protection is active (secure default). Pass `--disable-dns-rebinding-protection` to opt out when running behind a reverse proxy or port-forwarding tunnel. Auth is enforced separately via OAuth/PAT regardless.

**Tool & Prompt API Changes**
- Added `structured_output=False` to `mcp.add_tool()` to opt out of the SDK 1.14+ auto-detection that treats dict-returning tools as structured output, preserving backward-compatible result serialization.
- Fixed `Prompt.from_function()` call to use explicit `name=`/`title=`/`description=` kwargs instead of positional args (signature changed in SDK 1.14+).

### Compatibility & Testing

- Added `hasattr()` guard for `AppStatus.should_exit_event` (removed in sse-starlette 3.x).
- Updated 5 test files to handle the new result structure: migrated from `structuredContent` to `json.loads(result.content[0].text)` for tool result assertions.
- Fixed pre-existing assertion bugs in `test_allow_dml_without_ld` and `test_allow_dml_enabled_by_ld` that were masked by the old API (DML rejection returns `{"error": "..."}` at top level, not nested under a "result" key).
- Updated `test_meta_tools_disabled_at_invocation` for corrected `CallToolResult` access pattern (`result[0].text` instead of `result[0][0].text`).

### Documentation & Accessibility

Added `AGENTS.md` referencing `CLAUDE.md` in order to allow broader accessibility to both Claude and Codex agents. Created `AGENTS.md` as a reference file pointing to `CLAUDE.md` (via instruction syntax recommended by Codex) so that Codex can automatically discover project-specific instructions while maintaining the unified instruction set in `CLAUDE.md`.

## Test Plan

- [x] `uv run pytest tests/` — 394 passed, 4 skipped, 0 failures
- [x] Streaming HTTP smoke test: `curl http://127.0.0.1:8100/healthz` → 200 OK
- [x] Stdio smoke test: `initialize` RPC returns `serverInfo.version: 1.27.1`
- [x] All 13 required `mcp.*` imports verified working
- [x] `uv.lock` resolves `mcp==1.27.1`, `starlette==1.0.0`
- [x] `test_tool_pat[project_id]` — path-rewrite fix verified; `/mcp/{project_id}/` paths correctly rewrites to `/mcp` internally
- [x] DNS rebinding protection: default behavior (flag absent) — foreign Host headers return 421; with `--disable-dns-rebinding-protection` — foreign Host headers reach auth layer and return 401 Unauthorized
- [x] E2E tests: auth 401 passthrough, metrics, LaunchDarkly feature flags, OAuth discovery

---

🤖 Generated with [Claude Sonnet](https://claude.com/claude-code)


src/dremioai/servers/mcp.py
```diff
@@ -441,6 +442,14 @@ def init(
...
     if transport == Transports.streamable_http:
         opts["stateless_http"] = True
+        # SDK 1.14+ auto-enables DNS rebinding protection when bound to
+        # localhost, which rejects any Host header that isn't 127.0.0.1
+        # /localhost/::1. That breaks reverse-proxy and tunneled access
+        # patterns. Auth is enforced separately via OAuth/PAT, so disable
+        # this here to preserve pre-1.14 behavior.
+        opts["transport_security"] = TransportSecuritySettings(
+            enable_dns_rebinding_protection=False
+        )
```
ssaumitra: @aniket-s-kulkarni This will primarily apply for local debugging. Is it okay to keep it enabled globally or should we add an app configuration that will selectively enable this?

src/dremioai/tools/tools.py
```diff
@@ -134,30 +133,39 @@ def _df_to_json_records(df: pd.DataFrame) -> List[Dict[str, Any]]:
...
+                ProjectIdMiddleware.project_id_context.set(project_id)
+                FeatureFlagManager.set_project_id(project_id)
+                # Modify scope in-place so the MCP handler sees /mcp[/...].
+                # BaseHTTPMiddleware cannot do this: Starlette 1.0 call_next
+                # captures the original scope via closure and ignores any
+                # request.scope changes made in dispatch().
+                remaining = m.group(2).rstrip("/")
+                new_path = f"/mcp{remaining}" if remaining else "/mcp"
+                scope["path"] = new_path
+                scope["raw_path"] = new_path.encode("ascii")
```
aniket-s-kulkarni: why is this modification required?

src/dremioai/tools/tools.py
```diff
@@ -134,30 +133,39 @@ def _df_to_json_records(df: pd.DataFrame) -> List[Dict[str, Any]]:
...
+            else:
+                ProjectIdMiddleware.logger.debug(f"Path {path} doesn't match")
```
aniket-s-kulkarni: I know this is unrelated to the change, but lets change this a `.warning`

AGENTS.md
```diff
@@ -0,0 +1 @@
+Before working in this repository, read `CLAUDE.md` at the repository root and follow its instructions.
```
aniket-s-kulkarni, line 1: do we really need this .md file ?

src/dremioai/servers/mcp.py
aniket-s-kulkarni: @ssaumitra it would be good to understand what breaks (if anything) in prod with or without this.

src/dremioai/tools/tools.py
```diff
@@ -134,30 +133,39 @@ def _df_to_json_records(df: pd.DataFrame) -> List[Dict[str, Any]]:
     ]
 
 
-class ProjectIdMiddleware(BaseHTTPMiddleware):
+class ProjectIdMiddleware:
```
aniket-s-kulkarni, line 150: is `BaseHTTPMiddleware` gone?

pyproject.toml
```diff
@@ -14,14 +14,14 @@ dependencies = [
     "click>=8.1.8",
     "fastapi>=0.115.11",
     "launchdarkly-server-sdk>=9.15.0",
-    "mcp==1.10.0",
+    "mcp>=1.26.0,<2",
```
aniket-s-kulkarni: why `< 2` ?

src/dremioai/tools/tools.py
```diff
@@ -134,30 +133,39 @@ def _df_to_json_records(df: pd.DataFrame) -> List[Dict[str, Any]]:
...
-        if m := ProjectIdMiddleware.pat.search(request.url.path):
-            ProjectIdMiddleware.project_id_context.set(m.group(1))
-            FeatureFlagManager.set_project_id(m.group(1))
-        else:
-            ProjectIdMiddleware.logger.debug(
-                f"Path {request.url.path} ({request.url!r}) doesn't match"
-            )
-
-        return await call_next(request)
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
```
aniket-s-kulkarni, line 171: I would try and go back to `BaseHTTPMiddleware`, unless there is a problem we are solving.

aniket-s-kulkarni (changes requested):

src/dremioai/tools/tools.py
ssaumitra: I reverted this manually and brought back BaseHTTPMiddleware. I could see that  ProjectIdMiddleware completely stopped working. 

I did additional investigation with the help of Claude. I found that in mcp 1.12 (https://github.com/modelcontextprotocol/python-sdk/issues/732) a change was rolled out that made route matching strict (https://github.com/modelcontextprotocol/python-sdk/pull/1115). BaseHTTPMiddleware was removed so that `/mcp/{project_id}/` requests still reach to the MCP server and project ID parsing still works.

That being said, can you help me understand whether we have any dependency on `BaseHTTPMiddleware` ? If yes, I can investigate how can I bring it back.

ssaumitra: Also posting AI generated summary in a separate comment.

src/dremioai/tools/tools.py
ssaumitra: **Root cause: an SDK routing change broke the project-scoped endpoints, and the old `BaseHTTPMiddleware` couldn't fix it.**

**What changed in the SDK.** Up to mcp 1.10.0, FastMCP mounted the streamable-HTTP app with a Starlette `Mount`:
```python
Mount(self.settings.streamable_http_path, app=handle_streamable_http)
```
`Mount` is a **prefix** match, so requests like `/mcp/<project_id>/...` still resolved to the handler. In [modelcontextprotocol/python-sdk#1115](https://github.com/modelcontextprotocol/python-sdk/pull/1115) ("Update routing for streamable HTTP to avoid 307 redirect", commit `fd11eca`, first released in **v1.12.0**) this was switched to an exact `Route`:
```python
Route(self.settings.streamable_http_path, endpoint=streamable_http_app)
```
A `Route` is an **exact** path match. After the upgrade, `/mcp` works but `/mcp/<project_id>` returns **404** — the project-scoped endpoints silently stopped routing.

**Why the middleware had to be rewritten.** `ProjectIdMiddleware` already parses the `project_id` out of the path. The fix is to also **rewrite the request path** to plain `/mcp[/...]` so the now-exact `Route` matches again. A `BaseHTTPMiddleware.dispatch()` can't do that: in Starlette 1.0, `call_next` captures the original ASGI `scope` via closure and ignores any `request.scope` mutation made inside `dispatch()`. The only reliable place to rewrite `scope["path"]`/`scope["raw_path"]` before routing is **raw ASGI middleware**, which is why the class moved from `BaseHTTPMiddleware` to a plain `__call__(scope, receive, send)`.

So this isn't a stylistic refactor — it's the minimal change required to keep `/mcp/<project_id>` working after the `Mount`→`Route` change in SDK v1.12.0.

AGENTS.md
ssaumitra: I want to use Codex for rolling out PRs. Codex told me that it cannot use CLAUDE.md and presented three options

1. Put the actual instructions directly in AGENTS.md, or
2. Make AGENTS.md a symlink/copy of CLAUDE.md, if your environment preserves that, or
3. Add an explicit instruction like: Before working, read CLAUDE.md and follow it.

I opted for 3 over 2 because dremio-mcp is oss and some users might want to use windows for their development and symlink might create weird artifacts there.

src/dremioai/servers/mcp.py
ssaumitra: Check this https://github.com/dremio/dremio-mcp/pull/115#discussion_r3314265102

This change is mainly to make sure that local debugging setup (mcp server running locally, DNS rebound when forwarding the port, LLM using port-forwarded MCP). We can add this behind a flag or support key if we want.

src/dremioai/tools/tools.py
ssaumitra: No. This should answer it 
https://github.com/dremio/dremio-mcp/pull/115#discussion_r3325779124 and https://github.com/dremio/dremio-mcp/pull/115#discussion_r3325794715

pyproject.toml
ssaumitra: Ticket description for https://dremio.atlassian.net/browse/DX-121841 had an explicit instruction to add `< 2`. 

My initial understanding after reading the ticket description was that during the major upgrades, MCP library can roll out major changes during the major version upgrade that we should explicitly test before releasing. 

However, if this is not a case and we are confident to make MCP version upgrades automatically, then we can remove this.

src/dremioai/tools/tools.py
ssaumitra: Changing now

src/dremioai/tools/tools.py
aniket-s-kulkarni: Thanks. Makes sense.

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Yes lets add this behind a command line argument.

src/dremioai/tools/tools.py
aniket-s-kulkarni: All set

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Lets add this behind a command line option rather than a settings.py setting

AGENTS.md
aniket-s-kulkarni: with dap do we need either

src/dremioai/servers/mcp.py
aniket-s-kulkarni: @ssaumitra basically an option like `--enable-dns-rebinding` that is passed into this and enables this code.

pyproject.toml
aniket-s-kulkarni: @ssaumitra gotcha. There are more changes when going to the latest. We should go to the latest, could you create a new DX to go to the latest

src/dremioai/servers/mcp.py
ssaumitra: Makes sense. I will roll the change out rightaway.

AGENTS.md
ssaumitra: dap and CLAUDE.md / AGENTS.md solve different set of problems.

CLAUDE.md contains an overview about the project (how code is organized, dev guidelines, etc) that are specific to a given project. This file is loaded for any code prompt and LLM gets basic context about the project without spending tokens in every single session.
Source https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md

dap contains commonly used skills which list the common dev processes. Every single skill is not loaded in the every single LLM new session. Also skills do not contain repository specific information (e.g. how code is organized in each repository)
Source https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview

Also note that DAP is closed source whereas MCP is opensource. If an OSS contributor outside of Dremio would like to use AI for contributing to MCP, AGENTS.md and CLAUDE.md will aid them whereas dap will not.

pyproject.toml
ssaumitra: Probably some confusion.

MCP latest version right now is 1.27.2. https://pypi.org/project/mcp/#history Our tag numbers do resolve and include 1.27.2. So once this PR goes in, we will use the latest. 

After MCP 2.0 releases in the future, we can see whether there is any compatibility issue and based on that, we can plan for the upgrade. Are you okay with that?

src/dremioai/servers/mcp.py
ssaumitra: Done. Please take another look.

src/dremioai/tools/tools.py
```diff
@@ -134,30 +133,39 @@ def _df_to_json_records(df: pd.DataFrame) -> List[Dict[str, Any]]:
...
+                ProjectIdMiddleware.project_id_context.set(project_id)
+                FeatureFlagManager.set_project_id(project_id)
+                # Modify scope in-place so the MCP handler sees /mcp[/...].
+                # BaseHTTPMiddleware cannot do this: Starlette 1.0 call_next
+                # captures the original scope via closure and ignores any
+                # request.scope changes made in dispatch().
+                remaining = m.group(2).rstrip("/")
+                new_path = f"/mcp{remaining}" if remaining else "/mcp"
+                scope["path"] = new_path
+                scope["raw_path"] = new_path.encode("ascii")
```
aniket-s-kulkarni: - why does `raw_path` need to be set in addition to `path`
- is `encode()` really necessary

src/dremioai/servers/mcp.py
```diff
@@ -268,9 +269,18 @@ def protected_resource_path_from_request(
 def build_resource_metadata_url(
     request: Request, resource_path: str | None = None
 ) -> str:
+    if resource_path is None:
+        # ProjectIdMiddleware rewrites /mcp/{project_id}[/...] to /mcp[/...] in-place,
+        # so request.url.path has already lost the project_id by the time this is called.
+        # Reconstruct the original path from the context var when one is present.
+        if project_id := ProjectIdMiddleware.get_project_id():
+            current = request.url.path  # "/mcp" or "/mcp/messages"
```
aniket-s-kulkarni: its better to be explicit

- should we change `ProjectIdMiddle` to have a `get_remaining()` where that middlware also stores the trailing portion if any
- then this code becomes

```python
if project_id := ...:
    resource_path = f"/mcp/{project_id}/{ProjectIdMiddleware.get_remaining()}"
else:
    log.warning(....)
    ....
```

aniket-s-kulkarni (commented):
Looking good, just two small comments

ssaumitra:
I clicked "request review" by mistake before reading through your two new comments. Please ignore. 

I am going through your new feedbacks now.

src/dremioai/tools/tools.py
ssaumitra: I checked Starlette's codebase and did not find any evidence for `scope["raw_path"]`'s necessity. I locally tested removing this and everything worked correctly. I have now removed the `raw_path` logic..

src/dremioai/servers/mcp.py
ssaumitra: Done

aniket-s-kulkarni:
@ssaumitra in addition here are more notes

## MCP SDK upgrade notes (1.10 → latest `MCPServer`)

A review of the SDK changes relevant to this PR — summarising what's shifted and what needs retesting.

---

### Breaking import change

`from mcp.server.fastmcp import FastMCP` is replaced by `from mcp.server.mcpserver.server import MCPServer`.

The `FastMCPServerWithAuthToken` subclass pattern (overriding `streamable_http_app()` to stack auth middleware) is now replaced by passing `token_verifier=` directly to the `MCPServer` constructor.

---

### Auth middleware

`AuthenticationMiddleware`, `AuthContextMiddleware`, and `RequireAuthWithWWWAuthenticateMiddleware` are no longer manually stacked — the SDK wires these internally when `token_verifier` is provided.

**Please verify:**
- JWKS verification still works
- Token expiry buffer logic is respected
- `WWW-Authenticate` header is present on 401 responses

---

### `/.well-known/oauth-protected-resource` custom route

This is now auto-served by the SDK when `auth=AuthSettings(resource_server_url=...)` is set. If the custom route registration was removed as part of this PR, confirm the metadata document is still reachable and returns the correct payload.

---

### `streamable_http_app()` signature

The override no longer returns a mutable Starlette app — all options are passed as kwargs. Confirm `stateless_http=True` and `ProjectIdMiddleware` injection still work correctly.

---

### `add_tool()` / tool registration

Signature gains `title`, `meta`, `structured_output`. No breaking change to existing calls, but **plain dict returns from `invoke()` now also populate `structured_content` automatically** — verify tool responses look correct on the client side.

---

### Resource registration

`FunctionResource(uri=..., fn=...)` construction may need updating to `FunctionResource.from_function(...)` or the `@mcp.resource(...)` decorator pattern. Check that the `dremio://hints` resource is still served correctly.

---

### Retest checklist

Please confirm the following scenarios work end-to-end:

- [ ] stdio transport (no auth)
- [ ] streamable-http with a valid JWT
- [ ] streamable-http with an expired JWT → expect `401` + `WWW-Authenticate` header
- [ ] OAuth protected resource metadata endpoint (`/.well-known/oauth-protected-resource/mcp`)
- [ ] At least one tool call round-trip to confirm `structured_content` behaviour

ssaumitra:
I investigated into this
> Breaking import change
> from mcp.server.fastmcp import FastMCP is replaced by from mcp.server.mcpserver.server import MCPServer.
> The FastMCPServerWithAuthToken subclass pattern (overriding streamable_http_app() to stack auth middleware) is > now replaced by passing token_verifier= directly to the MCPServer constructor.

I see that the breaking change indeed exists in fastMCP's main branch. But v1.27.x still uses the good old FastMCP 
Source https://github.com/modelcontextprotocol/python-sdk/blob/v1.27.2/src/mcp/server/fastmcp/__init__.py#L7
FastMCP to MCPServer replacement was rolled out in https://github.com/modelcontextprotocol/python-sdk/pull/1951 but this not a part of 1.27.x release. Based on what I understand from their git repo, they are using `origin/main` for tracking the future 2.x release (where FastMCP to MCPServer rename will take place) and `origin/v1.x` branch is referring to 1.x releases.

>Resource registration
> FunctionResource(uri=..., fn=...) construction may need updating to FunctionResource.from_function(...) or the 
> @mcp.resource(...) decorator pattern. Check that the dremio://hints resource is still served correctly.

For FunctionResource as well, I do not see a breaking change in v1.x branch. It's however nice if we use `FunctionResource.from_function` . I am replacing the function call accordingly.

src/dremioai/servers/mcp.py
```diff
@@ -43,6 +43,7 @@
 from mcp.server.auth.middleware.bearer_auth import BearerAuthBackend
 from mcp.server.auth.provider import AccessToken, TokenVerifier
 from mcp.server.fastmcp import FastMCP
```
aniket-s-kulkarni, line 45: @ssaumitra how is this working? https://github.com/modelcontextprotocol/python-sdk/tree/main/src/mcp/server doesn't have FastMCP so it shouldn't work.

ssaumitra: See this https://github.com/dremio/dremio-mcp/pull/115#issuecomment-4617911768 . 

Based on my analysis, MCP 1.x releases are coming from origin/v1.x branch in the MCP repo. FastMCP to MCPServer rename has taken place in MCP repo's main branch (which will hopefully get released in MCP 2.x) 

If we inspect v1.x branch in the MCP repo (from where 1.x releases are coming out), we can see that FastMCP class exists and MCPServer class does not exist. In my previous comment, I included code reference to v1.27.2 (latest MCP release) where FastMCP class exists and MCPServer class does not exist.

AI's suggestion is therefore a red herring.

aniket-s-kulkarni: @ssaumitra fair enough. Ok so changes look good, have you been able to test it with the changes e2e?

ssaumitra: Yes, I tested with streamable HTTP CLI as well as connecting with chatgpt. Both work correctly.

aniket-s-kulkarni (approved):

PR #116, opened 2026-06-03
DX-122122: Support request-scoped dynamic MCP tools without FastMCP global tool cache

aniket-s-kulkarni:
### Core: request-scoped dynamic MCP tools (`src/dremioai/`)

- Override `list_tools()` and `call_tool()` in `FastMCPServerWithAuthToken` to merge static + request-scoped remote Dremio tools into a single `tools/list` response — LLMs no longer need a prior `DiscoverDynamicTools` call.
- Remote tools fetched via `@secured`-decorated `_list_remote_tools` / `_invoke_remote_tool` helpers, injecting request-scoped `PAT` + `project_id` without reimplementing `@secured`.
- `call_tool` delegates static tools to `super()`, routes remote tools directly through `_invoke_remote_tool` (bypasses `_tool_cache`), returns `result.model_dump(exclude_none=True)`.
- `DiscoverDynamicTools` / `CallDynamicTool` moved to new `ToolType.DYNAMIC_REMOTE_TOOLS` flag — no longer registered in default `FOR_DATA_PATTERNS` / `FOR_SELF` modes.
- New `dremioai ai-tools list` / `ai-tools call` CLI commands (`src/dremioai/api/cli/ai_tools.py`).

### Semantic layer tools

- `SearchMetrics(query)` and `GetTableRelationships(path: List[str])` — typed MCP tool classes (`FOR_DATA_PATTERNS | EXPERIMENTAL`) that call `invoke_tool("searchMetrics", ...)` / `invoke_tool("getTableRelationShips", ...)` respectively; gated on `dremio.enable_semantic_layer`.
- `SearchTableAndViews` gains:
  - `topN: Optional[int] = None` — defaults to `dremio.search_topN` (default **10**), capped to that limit; communicates non-exhaustive nature to the LLM.
  - `get_description()` override — when `enable_semantic_layer` is true, calls `ai_tools.get_semantic_layer_tool_descriptions()` at server-init time to embed live server-supplied descriptions for `searchMetrics` and `getTableRelationships` in the tool description.
  - Parallel `getTableRelationShips` enrichment per TABLE/VIEW result → added as `"relationships"` key (parsed via typed `TableRelationship` model).
  - Parallel `searchMetrics` call → metric results appended with `"result_type": "METRIC"` (parsed via typed `Metric` / `MetricsSearchResult` models).

### Typed semantic-layer wire models (`src/dremioai/api/dremio/ai_tools.py`)

- `TableRelationship` / `RelationshipUsage` — mirrors Java `ImmutableTableRelationship` (camelCase aliases).
- `Metric` / `MetricUsage` / `MetricColumn` / `MetricsSearchResult` — mirrors Java `ImmutableMetric`; handles `{"errorMessage": ..., "metrics": [...]}` envelope.
- `get_semantic_layer_tool_descriptions()` async helper — fetches descriptions for `searchMetrics` / `getTableRelationships` from the AI tool registry.

### Settings (`src/dremioai/config/settings.py`)

- `dremio.enable_semantic_layer` — `bool`, default `False`; `RuntimeMutable` (LaunchDarkly hot-toggle).
- `dremio.search_topN` — `int`, default **10**; caps `SearchTableAndViews` result count.
- `dremio.enable_remote_tools` — existing flag; gates remote tool passthrough in `list_tools` / `call_tool`.

### Streamable HTTP CLI (`tests/stremable_http_cli.py`)

- Token resolution: `--token` → `~/.config/dremioai/.auth.yaml` (keyed by normalized MCP URL) → OAuth PKCE flow; token + refresh-token cached on success.
- `TokenExpiredError` + `except* httpx.HTTPStatusError` — uses Python 3.11 `except*` (PEP 654) to unwrap `HTTPStatusError` from the `ExceptionGroup` emitted by `streamablehttp_client`'s internal `TaskGroup`; retries once after refresh.
- `AuthCache` / `AuthCacheStore` (`RootModel[dict[str, AuthCache]]`) — Pydantic models; cache keyed by `_normalize_url()` (adds `https://`, strips trailing `/`).
- `@with_auth` decorator on `cli list-tools` / `cli call-tool` — resolves token and handles refresh transparently.
- `auth check` → delegates cache write to `auth cache-update`; `auth cache-show` displays a redacted table.
- `cli list-tools` renders a Rich table with tool descriptions as Markdown.
- `--redirect-port` / `--redirect-path` on `cli list-tools` and `cli call-tool` for OAuth PKCE callbacks.

## Test plan

- [ ] `uv run pytest tests/servers/test_remote_tools.py -v` — 9 unit tests (merged listing, disabled flag, remote invocation, static routing, no cross-project cache, name collision, unknown tool, API error degradation, disabled-mode error dict)
- [ ] `uv run pytest tests/ -v` — 382 passed, 4 skipped, 0 failures
- [ ] `uv run python tests/stremable_http_cli.py cli list-tools --url https://mcp.qa.dremio.site/mcp/<proj> --client-id https://connectors.dremio.app/claude --redirect-path /Callback`
- [ ] `uv run python tests/stremable_http_cli.py auth check --url https://mcp.qa.dremio.site/mcp/<proj> --client-id https://connectors.dremio.app/claude --redirect-path /Callback` then `auth cache-show`
- [ ] E2E smoketest with `--check-remote-tools` flag against a Dremio instance with `enable_remote_tools: true`
- [ ] Semantic layer: enable `dremio.enable_semantic_layer: true` and verify `SearchTableAndViews` description includes `searchMetrics` / `getTableRelationships` descriptions, results include `"relationships"` and `"result_type": "METRIC"` entries

## JIRA

https://dremio.atlassian.net/browse/DX-122122

### DIFF
--- a/src/dremioai/api/dremio/ai_tools.py
+++ b/src/dremioai/api/dremio/ai_tools.py
@@ -62,6 +62,75 @@ class InvokeToolResponse(BaseModel):
         return self.result is None and self.error is None
 
 
+# ---------------------------------------------------------------------------
+# Semantic-layer response models
+# ---------------------------------------------------------------------------
+
+class RelationshipUsage(BaseModel):
+    """Historical usage count for a table relationship in a given month."""
+
+    year: str
+    month: str
+    count: int
+
+
+class TableRelationship(BaseModel):
+    """A join relationship between two tables, as returned by ``getTableRelationShips``."""
+
+    source_table_id: str = Field(alias="sourceTableId")
+    source_table_name: str = Field(alias="sourceTableName")
+    source_table_alias: Optional[str] = Field(None, alias="sourceTableAlias")
+    source_column_name: str = Field(alias="sourceColumnName")
+    target_table_id: str = Field(alias="targetTableId")
+    target_table_name: str = Field(alias="targetTableName")
+    target_table_alias: Optional[str] = Field(None, alias="targetTableAlias")
+    target_column_name: str = Field(alias="targetColumnName")
+    join_cardinality: Optional[str] = Field(None, alias="joinCardinality")
+    description: Optional[str] = None
+    usage_metrics: List[RelationshipUsage] = Field(default_factory=list, alias="usageMetrics")
+
+    model_config = ConfigDict(populate_by_name=True)
+
+
+class MetricUsage(BaseModel):
+    """Historical usage count for a metric in a given month."""
+
+    year: str
+    month: str
+    count: int
+
+
+class MetricColumn(BaseModel):
+    """A column referenced by a metric's SQL formula."""
+
+    table_name: Optional[str] = Field(None, alias="tableName")
+    column_name: str = Field(alias="columnName")
+
+    model_config = ConfigDict(populate_by_name=True)
+
+
+class Metric(BaseModel):
+    """A semantic-layer metric returned by ``searchMetrics``."""
+
+    name: str
+    description: Optional[str] = None
+    sql_formula: str = Field(alias="sqlFormula")
+    columns: List[MetricColumn] = Field(default_factory=list)
+    synonyms: List[str] = Field(default_factory=list)
+    usage_metrics: List[MetricUsage] = Field(default_factory=list, alias="usageMetrics")
+
+    model_config = ConfigDict(populate_by_name=True)
+
+
+class MetricsSearchResult(BaseModel):
+    """Top-level payload inside ``InvokeToolResponse.result`` for ``searchMetrics``."""
+
+    error_message: Optional[str] = Field(None, alias="errorMessage")
+    metrics: List[Metric] = Field(default_factory=list)
+
+    model_config = ConfigDict(populate_by_name=True)
+
+
 async def list_tools() -> ListToolsResponse:
     try:
         client = AsyncHttpClient()
@@ -93,3 +162,24 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
     except Exception:
         log.exception("Failed to invoke AI tool '%s'", tool_name)
         return InvokeToolResponse(error=f"Unexpected error invoking tool '{tool_name}'")
+
+
+_SEMANTIC_TOOL_NAMES = frozenset({"searchMetrics", "getTableRelationships"})
+
+
+async def get_semantic_layer_tool_descriptions() -> dict[str, str]:
+    """Return the descriptions for ``searchMetrics`` and ``getTableRelationships``
+    as advertised by Dremio's AI tool registry.
+
+    Used by :class:`SearchTableAndViews` to enrich its own tool description with
+    live, server-supplied text when ``enable_semantic_layer`` is configured.
+    Returns an empty dict when the registry is unavailable or returns an error.
+    """
+    response = await list_tools()
+    if response.error or not response.tools:
+        return {}
+    return {
+        t.name: (t.description or "")
+        for t in response.tools
+        if t.name in _SEMANTIC_TOOL_NAMES
+    }
--- a/src/dremioai/api/dremio/ai_tools.py
+++ b/src/dremioai/api/dremio/ai_tools.py
@@ -20,7 +20,7 @@ from urllib.parse import quote
 
 from aiohttp import ClientResponseError
 from dremioai.api.transport import DremioAsyncHttpClient as AsyncHttpClient
-from dremioai.config import settings
+from dremioai.config import settings, feature_flags
 from dremioai.log import logger
 
 log = logger(__name__)
@@ -42,6 +42,12 @@ class ListToolsResponse(BaseModel):
 
     def __bool__(self):
         return self.error is None
+
+
+def is_semantic_layer_enabled() -> bool:
+    return feature_flags.FeatureFlagManager.instance().get_flag(
+        "ai_semantic_layer.enabled", False
+    ) or settings.instance().dremio.get("enable_semantic_layer")
 
 
 class InvokeToolResponseResult(BaseModel):
--- a/src/dremioai/api/dremio/ai_tools.py
+++ b/src/dremioai/api/dremio/ai_tools.py
@@ -15,7 +15,7 @@
 #
 
 from pydantic import BaseModel, Field, ConfigDict
-from typing import Dict, List, Optional, Any
+from typing import Dict, List, Optional, Any, TypeVar
 from urllib.parse import quote
 
 from aiohttp import ClientResponseError
@@ -44,9 +44,38 @@ class ListToolsResponse(BaseModel):
         return self.error is None
 
 
+class InvokeToolResponseResult(BaseModel):
+    """Wraps the ``result`` object returned by an AI-tool invocation.
+
+    Extra fields from the server (e.g. ``columns``, ``rows``) are stored in
+    ``model_extra`` and exposed via dict-style access so callers can write
+    ``response.result["columns"]`` without changing their code.
+    """
+
+    model_config = ConfigDict(extra="allow", populate_by_name=True)
+
+    # Inner-result level error (Java uses "errorMessage" here)
+    error: Optional[str] = Field(default=None, alias="errorMessage")
+
+    def __bool__(self) -> bool:
+        return self.error is None
+
+    def __getitem__(self, key: str) -> Any:
+        """Allow dict-style read access to extra fields: ``result["columns"]``."""
+        if key in self.__class__.model_fields:
+            return getattr(self, key)
+        return self.model_extra[key]
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, dict):
+            return self.model_extra == other
+        return super().__eq__(other)
+
+
 class InvokeToolResponse(BaseModel):
 
-    result: Optional[Any] = None
+    result: Optional[InvokeToolResponseResult] = None
+    # Top-level error key from Dremio is "error" (not "errorMessage")
     error: Optional[str] = None
 
     def __bool__(self):
@@ -65,6 +94,7 @@ class InvokeToolResponse(BaseModel):
 # ---------------------------------------------------------------------------
 # Semantic-layer response models
 # ---------------------------------------------------------------------------
+
 
 class RelationshipUsage(BaseModel):
     """Historical usage count for a table relationship in a given month."""
@@ -87,9 +117,25 @@ class TableRelationship(BaseModel):
     target_column_name: str = Field(alias="targetColumnName")
     join_cardinality: Optional[str] = Field(None, alias="joinCardinality")
     description: Optional[str] = None
-    usage_metrics: List[RelationshipUsage] = Field(default_factory=list, alias="usageMetrics")
+    usage_metrics: List[RelationshipUsage] = Field(
+        default_factory=list, alias="usageMetrics"
+    )
 
     model_config = ConfigDict(populate_by_name=True)
+
+
+class TableRelationshipsResponse(BaseModel):
+    relationships: Optional[List[TableRelationship]] = Field(default_factory=list)
+    error: Optional[str] = Field(None, alias="errorMessage")
+
+    def __bool__(self):
+        return self.error is None
+
+    @property
+    def is_empty(self) -> bool:
+        return self.error is None and (
+            self.relationships is None or len(self.relationships) == 0
+        )
 
 
 class MetricUsage(BaseModel):
@@ -125,8 +171,15 @@ class Metric(BaseModel):
 class MetricsSearchResult(BaseModel):
     """Top-level payload inside ``InvokeToolResponse.result`` for ``searchMetrics``."""
 
-    error_message: Optional[str] = Field(None, alias="errorMessage")
+    error: Optional[str] = Field(None, alias="errorMessage")
     metrics: List[Metric] = Field(default_factory=list)
+
+    def __bool__(self):
+        return self.error is None
+
+    @property
+    def is_empty(self) -> bool:
+        return self.error is None and (self.metrics is None or len(self.metrics) == 0)
 
     model_config = ConfigDict(populate_by_name=True)
 
@@ -162,6 +215,36 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
     except Exception:
         log.exception("Failed to invoke AI tool '%s'", tool_name)
         return InvokeToolResponse(error=f"Unexpected error invoking tool '{tool_name}'")
+
+
+DeserT = TypeVar("DeserT", bound=BaseModel)
+
+
+async def invoke_tool_with_deser(
+    tool_name: str, args: Dict[str, Any], deser: DeserT
+) -> Optional[List[DeserT]]:
+    if resp := await invoke_tool(tool_name, args):
+        if resp.is_empty:
+            return []
+        try:
+            return deser.model_validate(resp.result.model_extra)
+        except:
+            log.exception(f"Failed to process {tool_name} ({args})")
+    else:
+        log.error(f"Failed {tool_name} for {args}: {resp}")
+    return None
+
+
+async def get_relationships(path: List[str]) -> Optional[TableRelationshipsResponse]:
+    return await invoke_tool_with_deser(
+        "getTableRelationships", {"path": path}, TableRelationshipsResponse
+    )
+
+
+async def get_metrics(query: str) -> Optional[List[Metric]]:
+    return await invoke_tool_with_deser(
+        "searchMetrics", {"query": query}, MetricsSearchResult
+    )
 
 
 _SEMANTIC_TOOL_NAMES = frozenset({"searchMetrics", "getTableRelationships"})
--- a/src/dremioai/servers/mcp.py
+++ b/src/dremioai/servers/mcp.py
@@ -628,7 +628,7 @@ def init(
                 else make_logged_invoke(tool.__name__, tool_instance.invoke)
             ),
             name=tool.__name__,
-            description=tool_instance.invoke.__doc__,
+            description=tool_instance.get_description(),
             annotations=ToolAnnotations(
                 readOnlyHint=not (is_sql_tool and allow_dml),
                 destructiveHint=bool(is_sql_tool and allow_dml),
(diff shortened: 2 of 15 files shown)


src/dremioai/servers/mcp.py
```diff
@@ -391,6 +394,55 @@ async def verify_token(self, token: str) -> AccessToken | None:
...
+                        inputSchema=rt.input_schema,
+                    )
+                )
+            return static_tools + remote_tools
+        except Exception:
+            self._logger.exception("error fetching remote tools")
+            return static_tools
+
+    async def call_tool(self, name: str, arguments: dict) -> list[ContentBlock] | dict[str, Any]:
+        static_names = {t.name for t in await super().list_tools()}
```
AfonsoP1nto, line 469: Every RPC calls super().list_tools() to build static_names, regardless of whether enable_remote_tools is even on. This is an O(N-tools) registry walk on every tool invocation with no caching. 

I think we should check self._tool_manager.tools for name in lookup rather than calling list_tools(), or move the enable_remote_tools guard to the top so static tool calls pay no extra cost.

src/dremioai/servers/mcp.py
```diff
@@ -391,6 +394,55 @@ async def verify_token(self, token: str) -> AccessToken | None:
...
+                    self._logger.warning(
+                        "remote tool name collides with static tool, skipping",
+                        name=rt.name,
+                    )
+                    continue
+                remote_tools.append(
+                    MCPTool(
+                        name=rt.name,
+                        description=rt.description,
+                        inputSchema=rt.input_schema,
```
AfonsoP1nto, line 458: Static tools all have ToolAnnotations set, but remote tools are constructed without annotations=. Any MCP client that uses annotations, including the existing smoketest's --check-annotations path will fail on the first remote tool in tools/list when enable_remote_tools=True. Shouldn't we add annotations here?

tests/servers/test_remote_tools.py
```diff
@@ -0,0 +1,198 @@
...
+            "tools": {"server_mode": ToolType.FOR_SELF.name},
+        }
+    )
+
+
+@pytest.fixture(autouse=True)
+def _base_settings(reset_settings_state):
+    settings.set_base_settings(_make_settings(enable_remote_tools=True))
+    yield
+    settings.reset_state_for_tests()
```
AfonsoP1nto, line 50: We take reset_settings_state as a parameter and also have settings.reset_state_for_tests() here, shouldn't we have just one approach instead of double resetting?

tests/stremable_http_cli.py
```diff
@@ -48,6 +52,231 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    1. *explicit_token* (``--token`` flag) — cache is bypassed entirely.
+    2. ``~/.config/dremioai/.auth.yaml`` ``token`` field.
+    3. Full OAuth PKCE flow (requires *client_id*); writes result to cache.
+
+    Returns ``(token, AuthCache)``.  The cache is empty when the caller supplied
+    an explicit token.
+    """
+    url = _normalize_url(url)
+    if explicit_token is not None:
+        return explicit_token, AuthCache()
```
AfonsoP1nto, line 176: When explicit_token is provided, _resolve_token returns (token, AuthCache()), so the client_id is discarded and the cache is never written. If the token then expires, _handle_token_expired receives an empty AuthCache, finds no refresh_token, and returns None to SystemExit(1). A user providing both --token and --client-id gets no indication that the token wasn't cached or that re-auth is needed. Shouldn't we include client_id in the return?

src/dremioai/servers/mcp.py
```diff
@@ -391,6 +394,55 @@ async def verify_token(self, token: str) -> AccessToken | None:
...
+    async def call_tool(self, name: str, arguments: dict) -> list[ContentBlock] | dict[str, Any]:
+        static_names = {t.name for t in await super().list_tools()}
+        if name in static_names:
+            return await super().call_tool(name, arguments)
+
+        if not settings.instance().dremio.get("enable_remote_tools"):
+            return {"error": f"Tool '{name}' not found (remote tools disabled)"}
+
+        result = await self._invoke_remote_tool(name, arguments)
+        return result.model_dump(exclude_none=True)
```
ssaumitra: 🔴 **Remote tool failures are reported to the client as successful calls.**

When `_invoke_remote_tool` fails (backend 500, tool-not-found, network error), `InvokeToolResponse.model_dump(exclude_none=True)` produces `{"error": "..."}` — a plain dict. The MCP low-level handler treats **any dict** as structured content and sets `isError=False`:

```python
# mcp/server/lowlevel/server.py:548-551
elif isinstance(results, dict):
    maybe_structured_content = cast(StructuredContent, results)
    unstructured_content = [types.TextContent(type="text", text=json.dumps(results, indent=2))]
# ... CallToolResult(..., isError=False)
```

Static tools that fail propagate `isError=True` (via `ToolError` → `_make_error_result`), but every failed remote call — including the "remote tools disabled" branch on line 441 and unknown-tool errors — comes back as `isError=False` with the error buried in `structuredContent`. A client/LLM that branches on `isError` (the canonical MCP failure signal) sees these as successes.

Consider returning a `CallToolResult(isError=True, ...)` (or raising) when `result.error` is set, so remote failures behave like static-tool failures.

src/dremioai/servers/mcp.py
ssaumitra, line 469: 🟡 **`super().list_tools()` is rebuilt on every `call_tool` invocation** just to test set membership — it constructs a fresh `MCPTool` for every registered tool each call. Cheaper to consult the tool manager directly, e.g. `self._tool_manager.get_tool(name) is not None` (or `name in self._tool_manager._tools`), avoiding the per-call O(n) object construction.

tests/stremable_http_cli.py
```diff
@@ -803,7 +1161,34 @@ async def _run_smoketests(
...
+            pp(f"{len(remote_tools)} remote tool(s) found: {[t.name for t in remote_tools]}")
+
+            # Call the first remote tool with no arguments to verify routing works
+            first_remote = remote_tools[0]
+            pp(f"Calling remote tool '{first_remote.name}'..", end=" ")
+            result = await session.call_tool(first_remote.name, {})
+            _assert(
+                result.structuredContent is not None or not result.isError,
+                f"Remote tool call failed: {result.content}",
+            )
```
ssaumitra, line 1225: 🟠 **This assertion can never fail for a remote tool call.**

Because `call_tool` always returns a dict for remote tools (see the `call_tool` override in `mcp.py`), `result.structuredContent` is **always non-None** — even for an error dict like `{"error": ...}`. The `or` short-circuits true and the assert passes regardless of whether the tool actually worked, so this E2E check reports `OK` even when the backend returns an error.

Assert on the real result shape instead, e.g. `not result.isError and "error" not in (result.structuredContent or {})`.

ssaumitra (commented):
## Code review — request-scoped dynamic MCP tools

Reviewed the branch against the installed `mcp` library's low-level result handling. Core design is sound — overriding `list_tools`/`call_tool`, the request-scoped `@secured` helpers reading per-request token/project from contextvars, and degrading to static tools on remote-listing failure are all the right calls.

Two findings I'd treat as blocking (#1 and #2): together they mean remote-tool failures are invisible to clients **and** the E2E smoketest meant to catch this can't fail. The rest are cleanup.

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 | Remote tool failures reported to client as `isError=False` (success) |
| 2 | 🟠 | `--check-remote-tools` smoketest assertion can never fail for a remote call |
| 3 | 🟠 | `mcp_client_session` misclassifies arbitrary errors as token expiry |
| 4 | 🟡 | Hardcoded static-tool allowlist in smoketest will drift |
| 5 | 🟡 | `call_tool` rebuilds the entire tool list on every invocation |
| 6 | 🟡 | `run_test` re-implements the `with_auth` resolve/refresh/retry inline |

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

tests/stremable_http_cli.py
```diff
@@ -304,34 +613,64 @@ async def mcp_client_session(
...
+    except httpx.HTTPStatusError as exc:
+        if exc.response.status_code == 401:
+            raise TokenExpiredError(str(exc)) from exc
+        raise
+    except Exception as exc:
+        # Fallback for non-httpx transports that signal auth failures via message text
+        msg = str(exc).lower()
+        if "401" in msg or "unauthorized" in msg:
+            raise TokenExpiredError(str(exc)) from exc
+        raise
```
ssaumitra: 🟠 **`mcp_client_session` can misclassify arbitrary errors as token expiry.**

Two issues:

1. `yield session` is inside the `try`, so exceptions raised by the **caller's `with` body** propagate back into this handler. Any unrelated failure whose message contains `"401"` or `"unauthorized"` (a SQL error, an assertion message, tool output text) is reclassified as `TokenExpiredError`, triggering a spurious refresh + full retry.
2. The substring match is the operative path because the MCP streamable client typically does not surface a bare `httpx.HTTPStatusError` on an init-time 401.

Prefer matching the transport's actual auth-error type, and narrow the `try` to the connect/initialize phase rather than wrapping the yielded body.

tests/stremable_http_cli.py
```diff
@@ -803,7 +1161,34 @@ async def _run_smoketests(
...
+        # 12. Remote tools: verify they appear in tools/list (optional)
+        # ------------------------------------------------------------------
+        if check_remote_tools:
+            pp("Checking remote tools in tools/list..", end=" ")
+            static_tool_names = {
+                "RunSqlQuery", "GetUsefulSystemTableNames", "GetSchemaOfTable",
+                "GetDescriptionOfTableOrSchema", "GetTableOrViewLineage",
+                "SearchTableAndViews", "DiscoverDynamicTools", "CallDynamicTool",
+            }
+            remote_tools = [t for t in tools_result.tools if t.name not in static_tool_names]
```
ssaumitra: 🟡 **Hardcoded static-tool allowlist will silently drift.**

This duplicates the server's tool registry as a literal. When a static tool is added or renamed it gets misclassified as "remote," and `len(remote_tools) > 0` can pass for the wrong reason (or fail spuriously). A more robust partition would query the server's static set (e.g. list tools with `enable_remote_tools` off, or expose the static names) rather than hardcoding.

tests/stremable_http_cli.py
```diff
@@ -482,28 +836,30 @@ async def run_test(
...
+        try:
+            await _do_remote(resolved)
+        except TokenExpiredError:
+            pp("[yellow]Token expired — refreshing...[/yellow]")
+            new_token = _handle_token_expired(url, cache)
+            if not new_token:
+                pp("[red]Token refresh failed.[/red]  Re-authenticate with --client-id or supply a new --token.")
+                raise SystemExit(1)
+            pp("[green]Token refreshed — retrying...[/green]")
+            await _do_remote(new_token)
```
ssaumitra: 🟡 **This re-implements the `with_auth` decorator's resolve/refresh/retry logic inline.**

It's a hand-copied version of `with_auth` (the `_resolve_token` → `except TokenExpiredError` → `_handle_token_expired` flow). Two copies of the same token lifecycle will diverge over time (the messages already differ slightly). Consider factoring the resolve+refresh+retry into one helper that both the decorator and `run_test` call.

tests/stremable_http_cli.py
aniket-s-kulkarni: The idea is that for testing, if you provide explicit token, that should never be stored. If you want ease of testing then don't provide explicit token and the cache will manage for you

maxlepikhin (approved):

tests/servers/test_remote_tools.py
aniket-s-kulkarni: Will take this up as techdebt

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Will take this up as tech debt - the MCP SDK update will also impact this,

src/dremioai/servers/mcp.py
aniket-s-kulkarni: Will take this up as tech debt - the backend should support this

src/dremioai/tools/tools.py
```diff
@@ -514,35 +532,210 @@ class SearchTableAndViews(Tools):
...
-                search.get_search_results(
-                    search.Search(query=query, filter=category), use_df=True
-                )
-                for category in (search.Category.TABLE, search.Category.VIEW)
-            ]
+        if not ai_tools.is_semantic_layer_enabled():
+            return {
+                "error": "Semantic layer is not enabled (dremio.enable_semantic_layer)."
+            }
+        result = await ai_tools.invoke_tool("getTableRelationShips", {"path": path})
```
AfonsoP1nto, line 727: The tool name is getTableRelationships, wont't this hit the wrong endpoint?

tests/stremable_http_cli.py
ssaumitra: Resolving for unblocking the release.

tests/stremable_http_cli.py
ssaumitra: Resolving for unblocking the release.

src/dremioai/servers/mcp.py
ssaumitra: Resolving for unblocking the release.

aniket-s-kulkarni: Will take this as techdebt

src/dremioai/tools/tools.py
```diff
@@ -514,35 +532,210 @@ class SearchTableAndViews(Tools):
...
+            return {
+                "error": "Semantic layer is not enabled (dremio.enable_semantic_layer)."
+            }
+        result = await ai_tools.invoke_tool("searchMetrics", {"query": query})
+        if result.error:
+            return {"error": result.error}
+        return (
+            result.result
+            if isinstance(result.result, dict)
+            else {"results": result.result}
```
AfonsoP1nto, line 700: I think isinstance(result.result, dict) is always False, as result.result is InvokeToolResponseResult, a Pydantic model. The else branch always executes, returning {"results": InvokeToolResponseResult(...)}. The MCP low-level server at mcp/server/lowlevel/server.py:468 calls json.dumps(results, indent=2) directly on the returned dict, raising TypeError: Object of type InvokeToolResponseResult is not JSON serializable.

src/dremioai/tools/tools.py
```diff
@@ -514,35 +532,210 @@ class SearchTableAndViews(Tools):
...
+            return {
+                "error": "Semantic layer is not enabled (dremio.enable_semantic_layer)."
+            }
+        result = await ai_tools.invoke_tool("getTableRelationShips", {"path": path})
+        if result.error:
+            return {"error": result.error}
+        return (
+            result.result
+            if isinstance(result.result, dict)
+            else {"result": result.result}
```
AfonsoP1nto, line 733: This has the same issue as I described in the SearchMetrics comment.

src/dremioai/tools/tools.py
```diff
@@ -514,35 +532,210 @@ class SearchTableAndViews(Tools):
...
+                by_alias=True, exclude_none=True
+            )
+
+        return ret
+
+
+class SearchMetrics(Tools):
+    """Search for metrics in the Dremio semantic layer using a natural-language query."""
+
+    For: ClassVar[Annotated[ToolType, ToolType.DYNAMIC_REMOTE_TOOLS]]
```
AfonsoP1nto, line 674: SearchMetrics and GetTableRelationships are classified as DYNAMIC_REMOTE_TOOLS, do we want to exclude them from the FOR_DATA_PATTERNS mode? test_fastmcp_basic.py, lists them in the args dict for a FOR_DATA_PATTERNS test, so I am unsure what the intent is

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -93,3 +221,54 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
...
+async def invoke_tool_with_deser(
+    tool_name: str, args: Dict[str, Any], deser: DeserT
+) -> Optional[List[DeserT]]:
+    if resp := await invoke_tool(tool_name, args):
+        if resp.is_empty:
+            return []
+        try:
+            return deser.model_validate(resp.result.model_extra)
+        except:
+            log.exception(f"Failed to process {tool_name} ({args})")
```
AfonsoP1nto, line 238: I think the except: here catches BaseException, including asyncio.CancelledError. In an async context, swallowing task cancellation signals causes the coroutine to return None instead of propagating the cancellation, breaking cooperative scheduling.

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -93,3 +221,54 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
...
+
+
+async def invoke_tool_with_deser(
+    tool_name: str, args: Dict[str, Any], deser: DeserT
+) -> Optional[List[DeserT]]:
+    if resp := await invoke_tool(tool_name, args):
+        if resp.is_empty:
+            return []
+        try:
+            return deser.model_validate(resp.result.model_extra)
```
AfonsoP1nto, line 236: This returns a single DeserT object, but the supposed return type should be Optional[List[DeserT]]

src/dremioai/api/dremio/ai_tools.py
```diff
@@ -93,3 +221,54 @@ async def invoke_tool(tool_name: str, args: Dict[str, Any]) -> InvokeToolRespons
...
+
+async def get_relationships(path: List[str]) -> Optional[TableRelationshipsResponse]:
+    return await invoke_tool_with_deser(
+        "getTableRelationships", {"path": path}, TableRelationshipsResponse
+    )
+
+
+async def get_metrics(query: str) -> Optional[List[Metric]]:
+    return await invoke_tool_with_deser(
+        "searchMetrics", {"query": query}, MetricsSearchResult
```
AfonsoP1nto, line 252: get_metrics() is typed Optional[List[Metric]] but receives Optional[MetricsSearchResult]

AfonsoP1nto (commented):
flagged a few issues I might have found for future improvements

src/dremioai/servers/mcp.py
ssaumitra: Resolving for unblocking the release.

PR #117, opened 2026-06-08
DX-122383: Limit RunSqlQuery responses and add observability

aniket-s-kulkarni:
## Summary
- fetch RunSqlQuery result pages sequentially with bounded row limits
- remove the default pandas-backed MCP response path for RunSqlQuery
- add row/page/truncation/response-size metrics and large-result mock coverage

## Tracking
- Jira: DX-122383

## Testing
- UV_CACHE_DIR=/tmp/uv-cache RUNSQL_SOAK_ITERS=20 RUNSQL_SOAK_SAMPLE_EVERY=10 uv run pytest tests/api/dremio/test_sql.py tests/tools/test_output_validation.py tests/e2e/test_metrics_e2e.py tests/e2e/test_runsql_large_e2e.py tests/e2e/test_runsql_soak_e2e.py -s


ssaumitra (changes requested):
Same point as mentioned in the other PR.

The limit functionality already exists in the Dremio query engine and REST API. I request to use that instead.

If we use limit parameter in the Job Results API (https://docs.dremio.com/current/reference/api/job/job-results/) the limits will be applied at the query plan level. Query planner will optimize the query to read less number of Parquet files, the Dremio query execution cost will be lower and Dollar cost/compute cost for the end user will also be lower. On the other hand, if we use limit functionality in MCP server layer, Dremio query engine will run the full cost query. And later trim the output. So user will pay full dollar cost for running the entire query and receive truncated results.

Double implementation of the same functionality at two different services will also end up confusing the end users. The Maintenance overhead will also be higher.

We already call Job results API [over here](https://github.com/dremio/dremio-mcp/blob/bb6f6d7fc7f8758bf2675732ead9bbeed98f3116/src/dremioai/api/dremio/sql.py#L166-L171) so the MCP integration will also be straight forward.

aniket-s-kulkarni:
@ssaumitra

> The limit functionality already exists in the Dremio query engine and REST API. I request to use that instead.

Good point to check. The `limit` parameter on the Job Results API is for limiting the number of rows retrieved, and we already use that path today. In this implementation, `page_size = min(500, fetch_rows)` drives the `limit` we send when fetching `/job/{id}/results`, so the Job Results API limit is already part of the retrieval flow.

The MCP-side logic here is bounding how much of the already-retrieved result set we return to the MCP client, rather than replacing the Job Results API limit.

aniket-s-kulkarni:
Updated this PR to remove row-count guardrails from `RunSqlQuery` and keep only byte-size guardrails.

What changed:
- removed `dremio.max_result_rows`
- `run_query(..., with_guardrails=True)` now fetches all rows sequentially
- `RunSqlQuery` truncation now happens only on response byte size
- `GetFailedJobDetails` still opts out of guardrails explicitly via `with_guardrails=False`

Validation:
- full test suite passed locally: `414 passed, 5 skipped`

src/dremioai/tools/tools.py
```diff
@@ -384,30 +453,97 @@ def ensure_query_allowed(s: str):
...
+            )
         try:
-            query = f"/* dremioai: submitter={self.__class__.__name__} */\n{query}"
-            df = await sql.run_query(query=query, use_df=True)
-            return {"result": _df_to_json_records(df)}
+            tagged_query = f"/* dremioai: submitter={self.__class__.__name__} */\n{query}"
+            dremio_settings = settings.instance().dremio
+            max_bytes = dremio_settings.get("max_result_bytes")
+            if max_bytes is None:
+                max_bytes = 204_800
```
ssaumitra, line 487: I created a data table where the table contained 6 rows, 5 rows being small and manageable and one field in one of the row containing large amount of data (that crosses this limit). When I queried that table, neither row or column trimming logic from this function helped. What helped though, was when LLM hit into the error from L521 here, it tried figuring out which field is large and truncated it by itself.

Here's the chat thread https://chatgpt.com/share/e/6a306978-e3c4-8326-8952-39aae77d2e6a

Given that this whole truncation functionality can be executed by the least complex LLM model (5.5 instant), I would recommend letting LLM handle the truncation complexities. From MCP side, when the output is too large, we can just throw an error from L521. That way MCP logic can stay clean and simple. With clearly defined threshold on result size, we will get reliability and LLM will handle truncating / retrying smaller queries / telling users to reduce the data size.

What do you think?

ssaumitra: Clarified over chat. Resolving.

ssaumitra (approved):

PR #122, opened 2026-07-01
DX-123244: Upgrading vulnerable dependencies

ssaumitra:
Pin aiohttp, cryptography, pyjwt, python-multipart, starlette, urllib3, pydantic-settings and idna to exact CVE-fixed versions (==); re-lock. Clears 30 Trivy CVEs (7 HIGH / 14 MEDIUM / 9 LOW).

`tests/e2e/test_launchdarkly_e2e.py`: `_build_jwt` built an unsigned test token via `jwt.encode(key="", algorithm="HS256")`, relying on PyJWT silently accepting an empty HMAC key with a warning. PyJWT 2.13.0 (pulled in by the pin above) hardens this into a hard InvalidKeyError, so the helper now builds the token with `algorithm="none"` instead — still unsigned, and safe here since the only consumer (`mcp.py`'s `extract_org_id_from_jwt path`) decodes with `verify_signature=False`.


tests/e2e/test_launchdarkly_e2e.py
```diff
@@ -53,9 +52,7 @@ def _make_mock_ld_client(flag_values: dict):
 
 def _build_jwt(aud: str) -> str:
     """Build a minimal unsigned JWT with the given ``aud`` claim."""
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", message=".*HMAC key.*")
-        return jwt.encode({"aud": aud}, key="", algorithm="HS256")
+    return jwt.encode({"aud": aud}, key=None, algorithm="none")
```
aniket-s-kulkarni: why is this necessary? This was just a test that was supressing warnings. Hows this related to the CVEs

ssaumitra: At the time of implementing this, I investigated this. Basically with `return jwt.encode({"aud": aud}, key="", algorithm="HS256")` we were previously building unsigned JWT token. After upgrading PyJWT to 2.13.0, PyJWT throws `InvalidKeyError` for the cases like this.

That being said, your concern is valid about removal of warnings. I am bringing in the warning suppression part and keep `algorithm="None"` part.

aniket-s-kulkarni (approved):

PR #123, opened 2026-07-16
DX-124126: Switch base image to Alpine to clear CVE-2026-57432 and its siblings

ssaumitra:
The Trivy scan of the image flagged perl-base 5.40.1-6 for an integer overflow in S_measure_struct that lets a crafted pack/unpack template read heap memory out of bounds. perl-base is not a Python dependency of this project — it ships as an Essential package in the python:3.13-slim (Debian trixie) base image, unrelated to anything in pyproject.toml/uv.lock. There is no fixed Debian version to upgrade to either: `apt-cache policy perl-base` in the base image shows 5.40.1-6 is already the newest candidate in the trixie repos.

Rather than purge just perl-base, a full Trivy scan of the base image showed it's one of many: python:3.13-slim (debian 13.6) carries 168 OS CVEs (3 CRITICAL, 19 HIGH), spread across perl-base, util-linux and friends (bsdutils, mount, login, libblkid1, libmount1, libuuid1, liblastlog2-2), ncurses, and gzip — none with fixed versions available yet. Purging packages one Jira ticket at a time doesn't scale against that queue, so this switches the base image to
python:3.13.14-alpine3.24, which Trivy reports as 0 OS vulnerabilities (HIGH/CRITICAL or otherwise).

Verified before making the switch:
- All 80 locked runtime dependencies (including native-extension packages: pandas, numpy, cryptography, pydantic-core, orjson, yarl) have musllinux wheels: `pip install --only-binary=:all: --dry-run` against the exported requirements.txt resolves cleanly on python:3.13.14-alpine3.24, so nothing needs to compile from source.
- The full multi-stage build succeeds unchanged apart from two Alpine/musl adjustments: `useradd` -> `adduser -D` (no useradd on busybox), and `pip install` gained `--only-binary=:all:` so a future dependency without a musllinux wheel fails the build loudly instead of silently falling back to a source build.
- The built image runs `dremio-mcp-server --help`/`run` correctly under the non-root appuser, and importing every native-extension dependency (pandas, numpy, cryptography, pydantic_core, aiohttp, orjson, PyJWT) succeeds.
- A Trivy scan of the resulting image confirms 0 OS-level vulnerabilities, versus 22 HIGH/CRITICAL on the trixie-based image.
- Base image size drops from 381 MB to 144 MB.

Known musl trade-offs, judged not applicable here: smaller default thread stack size (this is an asyncio-based I/O-bound service, not thread-heavy), and musl's malloc has historically been slower under heavy multithreaded allocation churn (same reasoning applies). Alpine's musl 1.2.4+ (3.18+) also fixed the old DNS-over-TCP truncation issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage
-FROM python:3.13-slim AS builder
+FROM python:3.13.14-alpine3.24 AS builder
 
 ENV UV_COMPILE_BYTECODE=1 \
     UV_LINK_MODE=copy
@@ -19,18 +19,20 @@ RUN uv build --wheel -o /dist
 
 
 # Runtime stage
-FROM python:3.13-slim
+FROM python:3.13.14-alpine3.24
 
-# Create non-root user
-RUN useradd -m -u 1001 appuser
+# Create non-root user (Alpine/busybox: adduser, not useradd)
+RUN adduser -D -u 1001 appuser
 
 WORKDIR /app
 
 # Copy wheel from builder
 COPY --from=builder /dist/*.whl /dist/requirements.txt /tmp/
 
-# Install the wheel and dependencies
-RUN pip install --no-cache-dir -r /tmp/requirements.txt
+# Install the wheel and dependencies. --only-binary=:all: fails the build
+# loudly if a future dependency lacks a musllinux wheel, instead of silently
+# compiling from source (and needing a C toolchain we don't install here).
+RUN pip install --no-cache-dir --only-binary=:all: -r /tmp/requirements.txt
 RUN pip install --no-cache-dir /tmp/dremioai*.whl
 RUN rm /tmp/*.whl /tmp/requirements.txt
 


maxlepikhin (approved):
The dockerfile looks fine, alpine may not include some  shell commands if they are used in helm chart for example. Was this tested in daas just in case?

ssaumitra:
## Approach Comparison: `perl-base` Removal vs. Alpine Base Switch

During development, two approaches were evaluated to address CVE-2026-57432:

### Approach 1: Remove `perl-base` from python:3.13-slim
**What it does:**
- Purges perl-base:5.40.1-6 from the Debian trixie base image at runtime
- Uses `apt-get remove --allow-remove-essential` to force removal of this Essential package
- Leaves the rest of the Debian trixie base image in place (glibc-based)

**Risk Assessment:**
- ✅ **Safe to remove:** No other installed packages depend on perl-base; only 3 maintainer scripts reference it, and only 1 (libpam-modules preinst) actually invokes it during package upgrade — a path never exercised by the Dockerfile
- ✅ **No runtime impact:** The app is pure Python; no perl is invoked during runtime
- ✅ **Verified:** Removal doesn't break apt/dpkg, ca-certificates, Python, or the dremio-mcp-server CLI

**CVE Coverage:**
- Fixes: **7 of 22 HIGH/CRITICAL** OS vulnerabilities (perl-base alone)
- Leaves unfixed: **15 HIGH/CRITICAL** vulnerabilities in util-linux (bsdutils, mount, libblkid1, libmount1, libuuid1), ncurses, gzip, login — none with available patches yet
- **Result:** Clears this ticket; 15 follow-up Trivy tickets queued behind it

**Pros:**
- Minimal Dockerfile changes
- Uses a commonly-deployed base (python:3.13-slim with glibc)
- Preserves bash, traditional Unix tooling
- Wheel/binary compatibility unchanged — all glibc-based wheels continue to work

**Cons:**
- Tactical fix for one CVE, not a solution to the broader base-image problem
- Leaves the image with 15 unfixed HIGH/CRITICAL findings from the OS layer
- Vulnerable to future apt/dpkg/maintainer-script assumptions about perl's presence

---

### Approach 2: Switch to python:3.13.14-alpine3.24 (Selected)
**What it does:**
- Replaces the base image entirely with Alpine Linux, which uses **musl libc** instead of **glibc** (used by Debian)
- Changes package manager from `apt` to `apk`; `useradd` becomes `adduser -D`
- Adds `--only-binary=:all:` to pip to prevent silent fallback to source builds

**Linux Binary/Wheel Compatibility: musl vs. glibc**

This is the critical change — understanding it is essential for evaluating the switch:

- **glibc (Debian/python:3.13-slim):** The standard C library on most Linux distributions. Python wheels distributed for "linux" typically mean glibc-linked binaries. These wheels contain compiled native extensions (C/C++/Rust code) dynamically linked against glibc at runtime.
  
- **musl (Alpine):** A lighter-weight, standards-compliant C library used by Alpine. Glibc wheels **do not work** on musl systems — they'll fail at runtime with `libc.so.6: no such file` errors because musl doesn't provide the same binary interface. Instead, packages must ship **musllinux** wheels (e.g., `pydantic_core-2.46.3-cp313-cp313-musllinux_1_2_aarch64.whl`).

**Verification before switching:**
- ✅ **All 80 locked dependencies ship musllinux wheels:** Verified via `pip install --only-binary=:all: --dry-run` on Alpine — pandas 3.0.2, numpy, cryptography, pydantic-core, orjson, yarl, and all others publish musllinux variants
- ✅ **No glibc-only packages:** If any dependency *only* shipped glibc wheels, the pip install would fail; the `--only-binary=:all:` flag prevents silent source compilation. Result: no silent build-time surprises
- ✅ **Build succeeds:** Multi-stage Dockerfile builds without error, only 2 musl adjustments needed
- ✅ **All native extensions load:** `import pandas, numpy, cryptography, pydantic_core, aiohttp, orjson, PyJWT` all succeed — these are heavy native-extension packages that must link against musl
- ✅ **CLI and imports work:** `dremio-mcp-server --help` and full dependency load verified in the built image
- ✅ **Zero OS vulnerabilities:** Trivy scan of the Alpine-based image confirms 0 HIGH/CRITICAL findings (0 total)

**CVE Coverage:**
- Fixes: **All 22 HIGH/CRITICAL** OS vulnerabilities from the base image
- Plus: **146 total OS CVEs** eliminated by switching to Alpine's minimal attack surface
- **Result:** Clears this ticket *and* the entire queue of future base-image tickets

**Pros:**
- Eliminates all 22 HIGH/CRITICAL base-image findings at once — sustainable long-term
- Smaller base image (144 MB vs. 381 MB) → ~260 MB smaller final image
- Alpine's minimal footprint reduces maintenance surface long-term
- Explicitly guards against future wheel-availability issues via `--only-binary=:all:`
- musl/musllinux wheels are becoming standard — major packages (pandas, numpy, cryptography, pydantic) all ship them

**Cons:**
- Wheel compatibility shift: Requires all dependencies to ship musllinux wheels (verified for all 80 locked deps, but a future dependency addition could break the build if it doesn't publish musllinux)
- musl vs. glibc behavioral differences in edge cases: different malloc, different locale handling, different DNS resolver behavior (though Alpine 3.18+ with musl 1.2.4+ fixed the DNS truncation issue that was the classic complaint)
- Less common in enterprises running legacy workloads (though increasingly standard for cloud-native services)
- Slightly different binary behavior under pathological thread stack conditions and high-contention malloc scenarios (not applicable to an asyncio-based I/O service)

---

### Decision Rationale
The Alpine approach was chosen because:
1. **Scope:** This addresses not just DX-124126, but the entire queue of base-image CVEs — the perl-base removal would fix this ticket, but 15 others (util-linux, ncurses, gzip, login) would arrive next.
2. **Verification:** All musl/musllinux assumptions were tested empirically; the switch is de-risked for this codebase's dependency set.
3. **Sustainability:** One base-image change beats a pattern of purging packages one Jira ticket at a time as Debian trixie CVEs surface.
4. **Wheel ecosystem:** All 80 locked dependencies publish musllinux wheels; the ecosystem has matured enough to make this decision now.
5. **Size:** A 40% smaller final image is a nice bonus for CI/CD and cluster deployments.
6. **Guardrails:** The `--only-binary=:all:` flag ensures future dependency additions fail loudly at build time if they lack musllinux support, rather than silently compiling from source.

### Fallback Plan
If Alpine encounters any unforeseen issue during testing, deployment, or operations, the perl-base removal approach remains a valid fallback. Both approaches are correct; Alpine is simply the more strategic long-term choice.

ssaumitra:
> The dockerfile looks fine, alpine may not include some shell commands if they are used in helm chart for example. Was this tested in daas just in case?

@maxlepikhin Good callout. I initially assumed that helm charts would not call shell commands from inside of the docker image so scoped my testing on whether MCP functionally works. But this is worth double checking. Do you suggest just launching daas cluster with the modified image or do you recommend testing it deeper?

maxlepikhin:
> > The dockerfile looks fine, alpine may not include some shell commands if they are used in helm chart for example. Was this tested in daas just in case?
> 
> @maxlepikhin Good callout. I initially assumed that helm charts would not call shell commands from inside of the docker image so scoped my testing on whether MCP functionally works. But this is worth double checking. Do you suggest just launching daas cluster with the modified image or do you recommend testing it deeper?

Yes, just daas. And even just inspecting the mcp helm chart in deployment repo.

ssaumitra:
I inspected mcp helm charts in the deployment repo and do not see anything that will get blocked because of alpine based image. I will merge this PR, create a new MCP image and run daas cluster just as an extra validation.

ssaumitra:
daas cluster creation and termination were successful 
https://test.jenkins-evo.drem.io/job/devops/job/daas/job/create/6337/
https://test.jenkins-evo.drem.io/job/devops/job/daas/job/delete/2180/

